PR #10, opened 2020-06-03
Feature/init frank

FDelporte:
needed to make minor modifications to be able to build


savageautomate:
Both forms of the following compile for me with no errors.

```
System.setProperty(org.slf4j.simple.SimpleLogger.DEFAULT_LOG_LEVEL_KEY, "INFO");
System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", "INFO");
```

FDelporte:
> Both forms of the following compile for me with no errors.
> 
> ```
> System.setProperty(org.slf4j.simple.SimpleLogger.DEFAULT_LOG_LEVEL_KEY, "INFO");
> System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", "INFO");
> ```

Strange indeed... But was a quick fix and the same form is now used everywhere

Issue #27, opened 2020-07-07
I2CRegister.write ignores length value


FDelporte:
As reported by @hackerjimbo 

There is an issue with I2CRegister.write (data, offset, length). It seems to ignore the length and fire the whole amount of data in one go.

Starting points for research
* https://github.com/Pi4J/pi4j-v2/blob/master/pi4j-core/src/main/java/com/pi4j/io/i2c/impl/DefaultI2CRegister.java#L84
* https://github.com/Pi4J/pi4j-v2/blob/75e1909006ef8fae8ad02d44f48096b368c634fd/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CRegisterDataWriter.java#L111

savageautomate:
If the error is not in the Pi4J core library, then its possible its either in the PiGpio Provider Plugin or the PiGpio JNI (wrapper) Library code.

- Are you using the native or remote (TCP socket) implementation of PiGpio?


 ---

See if you can trace the variables here and make sure all the input parameters are still correct and `length` has not been overwritten/modified.

### Pi4J PiGpio Provider Plugin
https://github.com/Pi4J/pi4j-v2/blob/a54d626c7a7d3750e31a63210c2abd1c983c04df/plugins/pi4j-plugin-pigpio/src/main/java/com/pi4j/plugin/pigpio/provider/i2c/PiGpioI2C.java#L117

```
    public int write(byte[] data, int offset, int length) throws IOException {
        Objects.checkFromIndexSize(offset, length, data.length);
        piGpio.i2cWriteDevice(this.handle, data, offset, length);
        return length;
    }
```

### Pi4J PiGpio JNI (Wrapper) Library
https://github.com/Pi4J/pi4j-v2/blob/a54d626c7a7d3750e31a63210c2abd1c983c04df/libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/PiGpio_I2C.java#L1428


Then .. depending on if using native PiGpio or remote via TCP  ...

**_TCP/SOCKET IMPL:_**
https://github.com/Pi4J/pi4j-v2/blob/a54d626c7a7d3750e31a63210c2abd1c983c04df/libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/impl/PiGpioSocketImpl.java#L1117
```
    public int i2cWriteDevice(int handle, byte[] data, int offset, int length) throws IOException {
[...]

PR #29, opened 2020-07-08
FIX: Issue #27; I2CRegister.write ignores length value

savageautomate:
applied changes per @hackerjimbo ref: https://github.com/Pi4J/pi4j-v2/issues/27#issuecomment-655337273 for Issue #27

### DIFF
--- a/libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/impl/PiGpioNativeImpl.java
+++ b/libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/impl/PiGpioNativeImpl.java
@@ -792,7 +792,7 @@ public class PiGpioNativeImpl extends PiGpioBase implements PiGpio {
      */
     @Override
     public int i2cOpen(int bus, int device, int flags) throws IOException {
-        logger.trace("[I2C::OPEN] -> Open I2C Bus [{}] and Device [{}]", bus, device);
+        logger.trace("[I2C::OPEN] -> Open I2C Bus [{}] and Device [{}]; FLAGS=[{}]", bus, device, flags);
         validateReady();
         validateI2cBus(bus);
         validateI2cDeviceAddress(device);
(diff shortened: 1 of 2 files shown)


hackerjimbo:
This is embarrasing! My git-fu isn't up to grabbing this branch. I tried following the instructions to “bring in the changes and test” but got an error on the git checkout (fatal: 'origin/issue/#27' is not a commit and a branch 'issue/#27' cannot be created from it). Any ideas? Apart from that I _really_ ought to learn more git.

savageautomate:
If you have made any modifications in your current branch, you will have to either revert them or `stash` them to get back to the unmodified `master` branch..  

```
git stash
```

Next do a git pull to pull in all the latest changes and new branches:
```
git pull
```
Next, to switch to the new branch, use the `checkout` command.
```
git checkout issue/#27
```

This will pull in the commits where I made the changes for the correct `length` check.  This however, will **not** pull the changes related to compiling for the ARMv6. 

Thanks, Robert

hackerjimbo:
That last command, `git checkout issue/#27`, gave me the error `error: pathspec 'issue/#27' did not match any file(s) known to git`. That's what confused me when I tried it before.

FDelporte:
@hackerjimbo one rule for git if all else fails... :-) 
delete your local copy completely and just check out again from 

`git clone https://github.com/Pi4J/pi4j-v2.git`

hackerjimbo:
Well I've got some good news, and I've got some bad news.

I ran it on a machine _without_ the attached device (too see what would happen). The good news is that there is no longer an error being raised do to the length.

The bad news is that even though the I2C write failed, the return value from the write function returns the number of bytes I wanted to write rather than the number actually written (or an exception).

hackerjimbo:
Oh, and some of the traces still refer to the size of the buffer and don't mention the offset and length. For example, `[main] TRACE com.pi4j.library.pigpio.impl.PiGpioNativeImpl - [I2C::WRITE] -> [0]; Register [64]; I2C Block [128 bytes]`.

savageautomate:
> The bad news is that even though the I2C write failed, the return value from the write function returns the number of bytes I wanted to write rather than the number actually written (or an exception).

I was not able to track down in code where this could happen.  It might need to be traced out.  Try out the latest commits (in the JAR I forward to you) and lets see if the issues are resolved.     Thanks, Robert

Issue #15, opened 2020-06-12
Use of signal handlers needs to be improved


FDelporte:
As reported by Jim Darby:
_If you're running as root (never a good idea) and using pigpio directly then pigpio steals all the signal handlers. The JVM doesn't like this as it uses them in running the JVM so it ends up with the program crashing. I'm working on this by getting pi4j-v2 to put back the signal handlers that the JVM uses._

hackerjimbo:
As a first step I printed out the signal handled at the very start of Java_com_pi4j_library_pigpio_internal_PIGPIO_gpioInitialise to see what it was life before gpioInitialise is called. It looks like quite a lot.

Then I looked at the code in pigpio and found something very interesting. There is actually a compile-time #define, EMBEDDED_IN_VM, which explicitly turns off pigpio's use of signals!

Furthermore, there is a run-time flag as well. The code looks like:

#ifndef EMBEDDED_IN_VM
   if (!(gpioCfg.internals & PI_CFG_NOSIGHANDLER))
      sigSetHandler();
#endif

So, we can avoid the entire problem and use the OS-supplied pigpio by setting that flag!

This looks like the footprints of someone being here before. I'm going to guess that it's Python. Now all we have to do is to figure out how to set that flag.

hackerjimbo:
Brace yourselves. Here's the fix (I hope). In the native C bindings we have the function `Java_com_pi4j_library_pigpio_internal_PIGPIO_gpioInitialise`. This contains the single line ` return gpioInitialise();`.

Immediately before that insert `gpioCfgSetInternals (gpioCfgGetInternals () | PI_CFG_NOSIGHANDLER);`

I can't actually test this as I can't build an armv6 native library right now, but I strongly believe it should work.

savageautomate:
Here is the compiled JAR with the [changes you suggested](https://github.com/Pi4J/pi4j-v2/commit/b3e40b7370d14d3149a98a3c15c8ee47257b813d) if you would like to test it out:
 - https://www.dropbox.com/s/8cm176mbknk2s0m/pi4j-library-pigpio-2.0-SNAPSHOT.jar

https://github.com/Pi4J/pi4j-v2/blob/b3e40b7370d14d3149a98a3c15c8ee47257b813d/libraries/pi4j-library-pigpio/src/main/native/com_pi4j_library_pigpio_internal_PIGPIO.c#L155-L166

savageautomate:
@hackerjimbo ,

[...]

PR #32, opened 2020-07-13
FIX: Issue #15; Prevent the PiGPIO library from internally handling s…

savageautomate:
…ignals and preventing signals from reaching the JVM.


FDelporte:
Great work @savageautomate @hackerjimbo! Seems the biggest improvements only need a one line change ;-)

Issue #18, opened 2020-06-12
Extend the Context with an ExecutorService


FDelporte:
Different approaches should be possible
* User can inject his own executor service into the Context 
* The Context needs to create its own ExecutorService if none is provided

This ExecutorService is needed for
* DigitalOut Blink
* DigitalOut Pulse

https://www.baeldung.com/java-executor-service-tutorial

tcfurrer:
Multi-threaded capability and implementations of the asynchronous APIs would be a terrific step forward!

I assume this would require some non-trivial changes to provide the necessary thread safety, correct?

FDelporte:
Indeed @tcfurrer will have some impact. I have been looking into this in the past.

It would all start with defining the ExecutorService in the context
https://github.com/Pi4J/pi4j-v2/blob/develop/pi4j-core/src/main/java/com/pi4j/context/Context.java

And could be used to handle for instance thread actions like pulsing a LED
https://github.com/Pi4J/pi4j-v2/blob/develop/pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java#L102

These are only a few first steps and to be further investigated where and how such multi-threading should be applied.

Here would be an ideal place to discuss this functionality.

stefanhaustein:
Why would we want to add blink capabilities to digital IO? Isn't that totally redundant with PWM?

stefanhaustein:
Looks like this was done anyways...

PR #46, opened 2021-01-18
#45 reworked to full junit 5 (removing mixed 4/5 tests)

FDelporte:


FDelporte:
A lot of changes, but actually no code changes, only removing junit4 as there was a mixture of 4 and 5 in the project. 

Code changes = change of parameters of assert methods where the label needs to be at the last position instead of first.

pi4j-test/src/test/java/com/pi4j/test/platform/ManualPlatformsTest.java
```diff
@@ -39,18 +42,18 @@
...
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.TestInstance.Lifecycle;
 
+@TestInstance(Lifecycle.PER_CLASS)
 public class ManualPlatformsTest {
 
     private Context pi4j;
 
-    @Before
+    @BeforeAll
```
eitch, line 56: This won't work, as @BeforeAll must be a static method:

https://junit.org/junit5/docs/current/user-guide/#writing-tests-annotations

pi4j-core/src/main/java/com/pi4j/runtime/impl/DefaultRuntime.java
```diff
@@ -70,7 +67,6 @@
     private final RuntimeProperties properties;
     private final List<Plugin> plugins = new ArrayList<>();
     private boolean isShutdown = false;
-    private final ExecutorService executor = Executors.newSingleThreadExecutor();
```
eitch, line 73: The changes here seem to not be required for the JUnit changes, no? Should this be in a different pull request? Or why did you remove the shutting down?

eitch (changes requested):
- The @BeforeAll must be changed, either to @BeforeEach, or the method must be static.
- The runtime changes shouldn't be in this pull request

pi4j-test/src/test/java/com/pi4j/test/platform/ManualPlatformsTest.java
FDelporte: yes and no ;-)
by adding @TestInstance(Lifecycle.PER_CLASS) to the class, it is not needed to be static. 
static would also break the reference to private Context pi4j

pi4j-core/src/main/java/com/pi4j/runtime/impl/DefaultRuntime.java
FDelporte: this slipped in while working (long time ago) on #18 and is still part of branch https://github.com/Pi4J/pi4j-v2/tree/feature/%2318. In that branch I also did the move to junit5 but extracted it into this separate branch so this could be merged independent of the work on the executorservice which actually still needs architectural investigation first.

pi4j-test/src/test/java/com/pi4j/test/platform/ManualPlatformsTest.java
eitch: oh my... you are right. Sorry.

eitch:
@FDelporte now there is conflict open. I guess after you rebase the feature branch with master then we can merge the pull request

FDelporte:
@eitch conflict is fixed

savageautomate (approved):

eitch (approved):

Issue #53, opened 2021-01-20
Coding best practice: do not use system.out but only logging


FDelporte:
Review all code to:
* remove commented log-configuration lines or reenable and use them 
* replace all system.out with log

savageautomate:
Agreed.

PR #54, opened 2021-01-20
#53 replace all system.out to logger

FDelporte:


FDelporte:
Again no real code changes, but replaced all system.out with logger

eitch (approved):

PR #101, opened 2021-05-10
Use current Thread's context ClassLoader when creating provider proxy…

### LINKED ISSUE #100: Better Support for Custom Plugin Class Loading Strategies
I'd like to compile my code against `pi4j-core` but allow the user to provide plugin jars (such as `pi4j-plugin-mock`) in a well-known directory at runtime. In order to do this, I need to use a custom ClassLoader to look for plugins in the plugins directory. This is typically done by setting the current Thread's context ClassLoader, but current PI4J v2 code is not respecting this ClassLoader in `com.pi4j.provider.impl.DefaultRuntimeProviders#add(Collection<T> provider)` (around line 267). When creating a provider proxy, it is instead using the ClassLoader which loaded the `com.pi4j.provider.Provider` class, which will be the Application's ClassLoader instance (which will not know how to find plugins in our special directory).

The simple fix is to change this provider proxy creation code from:

```java
var providerProxy = Proxy.newProxyInstance(
        Provider.class.getClassLoader(),
        ReflectionUtil.getAllInterfaces(providerInstance).toArray(new Class[]{}),
        handler);
```

to:

```java
var providerProxy = Proxy.newProxyInstance(
        Thread.currentThread().getContextClassLoader(),
        ReflectionUtil.getAllInterfaces(providerInstance).toArray(new Class[]{}),
        handler);
```

See this SSCCE for a more detailed example: https://github.com/joelspecht/pi4j-v2-classloader-sscce

joelspecht:
… instance

Fixes #100

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/provider/impl/DefaultRuntimeProviders.java
+++ b/pi4j-core/src/main/java/com/pi4j/provider/impl/DefaultRuntimeProviders.java
@@ -265,7 +265,7 @@ public class DefaultRuntimeProviders implements RuntimeProviders {
 
                 ProviderProxyHandler handler = new ProviderProxyHandler(runtime, providerInstance);
                 var providerProxy = Proxy.newProxyInstance(
-                        Provider.class.getClassLoader(),
+                        Thread.currentThread().getContextClassLoader(),
                         ReflectionUtil.getAllInterfaces(providerInstance).toArray(new Class[]{}),
                         handler);
 


FDelporte:
Thanks @joelspecht for your contribution! 
We are currently reviewing the plugin approach and I think this fits into that process, so requested a review from team members.

joelspecht:
Glad this might fit in - thanks for reviewing / considering it!

PR #123, opened 2021-08-31
OSGi -enhancement - Use bnd-maven-plugin for better osgi support

stbischof:
Signed-off-by: Stefan Bischof <stbischof@bipolis.org>


stbischof:
could you have a look on this?
Especially the way of auto-generating the module-info.class

What are the needs to build this as an module and not just as an simple jar?

FDelporte:
Removing the module-info.java files will break the creation of the java modules, so this won't be a good approach... unfortunately I don't have enough OSGI experience to know how to support both modules and OSGI.

@savageautomate @eitch you don't happen to have the right knowledge regarding this topic?

stbischof:
The module-info.java is automatically generated from the export and provideservice annotations.

eitch:
So i of course like the idea of supporting OSGi, but one thing that won't work now is the fixed versioning in the code. The annotations in the package-info.java files are hard coded and we use mvn versions:set which probably wouldn't change these files. 

Could you maybe find a solution to auto-generate the package-info files with the current version?

eitch:
@stbischof did you not find an idea on how to handle the package-info classes?

stbischof:
Nope. And i had a deeper look into your plugin structure. I think it is better for me to to that with eclipse kura.

FDelporte:
@stbischof thanks for the feedback! Kura seems indeed good solution for OSGi approach

PR #184, opened 2022-01-06
No temporary memory allocation in ByteBuffer APIs.

haumacher:
APIs using `ByteBuffer`s are very convenient - much better than (byte[], off, len) APIs. Fortunatelly, pi4j supports such APIs, but unfortunately, it is currently not possible to create memory-efficient code using these APIs. Internally, temporary byte arrays are allocated for communication (instead of using the `ByteBuffer`s internal array). What is the reason behind this design? When communicating with devices such as gyroscopes and accelerometers a high frequency of calls are neccessary to transfer data from the device. If each call allocates temporary memory, this produces a high pressure on the PIs GC, potentially resulting in GC pauses on the device. 

I'd like to suggest to avoid allocating temporary memory whenever possible. The change in this pull-request is one example of replacing temporary memory with the internal byte array of a buffer. If such a change is welcome, I could create a more complete change.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CRegisterDataReader.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CRegisterDataReader.java
@@ -133,22 +133,16 @@ public interface I2CRegisterDataReader {
             length = buffer.capacity()-offset;
         }
 
-        // create a temporary byte array to read in the length of data bytes
-        byte[] temp = new byte[length];
-        int actualLength = readRegister(register, temp, 0 ,length);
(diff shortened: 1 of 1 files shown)


FDelporte:
dear @haumacher is it OK if I assign the issue #185  to you?

haumacher:
Yes please.

FDelporte:
@haumacher I think you will need to add a comment to the use #185 itself before I can assign it to you...

haumacher:
Ok, I'll try looking into #185.

eitch (approved):
LGTM

PR #193, opened 2022-02-04
added gpioCfgClock() to be able to configure sample rate and peripheral

hagego:


eitch (approved):
looks good

FDelporte:
@hagego if you want to try out, there is already a snapshot version available as you can see on https://pi4j.com/download/

[pi4j-2.2.0-SNAPSHOT.zip (2.45 MiB, 2022-02-07 07:04)](https://github.com/Pi4J/download/raw/main/pi4j-2.2.0-SNAPSHOT.zip)

PR #198, opened 2022-02-26
issue#40 implement deregistration/shutdown for IO's

beulenbilly:
Let me know if it works for you or not


pi4j-core/src/main/java/com/pi4j/context/impl/DefaultContext.java
```diff
@@ -133,7 +135,23 @@ public Context shutdown() throws ShutdownException {
         // shutdown the runtime
         this.runtime.shutdown();
         return this;
-    }
+	}
+
+	@Override
+	public <T extends IO> T shutdown(T io) {
+		if (null == io) {
```
eitch, line 142: i would be defensive here, and throw an IllegalArgumentException

pi4j-core/src/main/java/com/pi4j/context/impl/DefaultContext.java
```diff
@@ -133,7 +135,23 @@ public Context shutdown() throws ShutdownException {
...
+			return null;
+		}
+		return shutdown(io.getId());
+	}
+
+	@Override
+	public <T extends IO> T shutdown(String id) {
+		if (!hasIO(id)) {
```
eitch, line 150: i would be defensive here, and throw an IllegalArgumentException

eitch: I take that back. Looking at the implementation, exceptions are already thrown, remove the null checks and just call remove on the registry

eitch (changes requested):

pi4j-core/src/main/java/com/pi4j/context/impl/DefaultContext.java
beulenbilly: Removed the checks, added some Java doc and removed shutdown(IO io) because:

1. all other interfaces accepting just the string-id
2. avoid null pointer handling if the IO is null.

eitch (approved):

Issue #197, opened 2022-02-24
Sensor provider and interface


taartspi:
Within pi4j-core I created a new IO type 'sensor'  I created the config and provider classes.

FDelporte:
Hey @taartspi, thanks for your work! Are you working in a fork of this project? In that case you can create a pull request from your project to this one.

taartspi:
Morning.   No I used the url and of the core repo. But I will recreate a fork and move the parts into that repo. Thanks sir

taartspi:
I forked the pi4j V2 and added my Sensor code.  That has now been pushed and I created a pull request for your viewing.     And these changes can be tested using my repo https://github.com/taartspi/Pi4J_V2-TemperatureSensor    Once we agree on what I was implementing in the core, I would like review/comments on the BMP280 work whether it could stand as a simple example of device implementation.

taartspi:
@FDelporte      Hello Frank.  Is there an expected timeframe when my pull request would be considered ?  I don't know how busy all of you are, but I prefer the pull not exceed 30 days and be deleted.  As I said in the details if the pull was accepted after review/updates, there is a BMP280 implementation I would like considered as a simple device example.  thanks   Hope you are well.   Tom

FDelporte:
sorry, for delay, too few hours in a day, will take a look ;-)

taartspi:
@FDelporte @savageautomate 
The changes to the core is the addition of a Sensor interface and Sensor providers. The provider assumes the sensor needs a bus and address but does not assume the state pertains to i2c or spi or any other interface. 

In my own repo i used these to implement the code specific to a BMP280

taartspi:
pull  #199 
Long ago we talked about a generic Sensor interface, this is a shot at creating that and its Config and Provider  classes.

Then in my own repo I did an implementation specific to BMP280 using these core changes.  If this is something you still want to discuss let me know, if no longer of interest I will delete the request to clear it off the books

Tom

FDelporte:
I see the use-case for a generic sensor interface. So these methods indeed fit in this approach

```
resetSensor
initSensor
```

What doesn't really "fit" into the generic approach of the V2 library, or specific methods like the following as they are coupled to a specific use-case and type of sensor. 

```
temperatureC()
temperatureF()
pressurePa() 
pressureIn() 
pressureMb()
```

I'm not sure yet how this could be made more generic, but thinking more of an approach with e.g.

```
value(SensorType type)
```

[...]

PR #199, opened 2022-02-26
Add new IO type and providers and interface  Sensor

taartspi:
This add support into the pi4j-core.  It is a new IO type  Sensor. also defines an interface Sensor, Config/builder and provider.

There exist https://github.com/taartspi/Pi4J_V2-TemperatureSensor      This implements an I2C device that uses a BMP280 sensor, requiring this branch.  This branch also documents connecting an Adafruit BMP280 to a Pi.


taartspi:
@FDelporte       Good day.   I don't see the proper way to connect this pull request to the issue 197.   You have a quick pointer on doing this ?        thanks Tom

FDelporte:
I'm not really sure about this PR. It brings in a specific use-case/device, something that we removed when moving from V1 to V2 to improve maintainability. What do you think @eitch @savageautomate ?

FDelporte:
Relates to #197

taartspi:
Determined this code change is too specific to only two sensor types.  Suggested this implementation could define only the generic interfaces of possible  sensor types. Suggested ENUM be used.

PR #215, opened 2022-04-28
Add java_doc to make debounce default visible on input gpio config bu…

taartspi:
In debugging missed interrupts from a chip I realized the GPIO input debounce default value was greater than the chips interrupt duration.  This is just an added line of javadoc so the default details are available when hovering over the debounce() method when :    
DigitalInput.newConfigBuilder(pi4j)   
      .debounce()


FDelporte:
Thanks!

PR #230, opened 2022-06-27
Issues/#117

FDelporte:
I propose to merge the current state of the extended LinuxFS support. That will allow us to easily test it with a SNAPSHOT version. What do you think @savageautomate and @eitch ?


eitch (approved):
It's a big change, i think we should just merge and retest everything

PR #234, opened 2022-08-11
The spi provider docs states mode 1 and 2 are not supported on the sec…

taartspi:
The spi provider docs states mode 1 and 2 are not supported on the secondary spi.  The test was not correct and it prevented mode 1 on spi0

The only testing I did was to validate I can communicate to my spi device mode 1 from spi0.

if there is other testing you have that I can/should run, point me to the tests.


taartspi:
Looking at the comments through the spi provider, I think there is more conflict between comment and code.

PR #246, opened 2022-10-12
since no unique except thrown with ELF32/ELF64 violation. Examine err…

taartspi:
…or message for wrong ELF class. If found in except data throw detailed message to try and inform the user of the error. Change in normal and pi4j.library.path set to library location.

### DIFF
--- a/libraries/pi4j-library-linuxfs/src/main/java/com/pi4j/library/linuxfs/util/NativeLibraryLoader.java
+++ b/libraries/pi4j-library-linuxfs/src/main/java/com/pi4j/library/linuxfs/util/NativeLibraryLoader.java
@@ -123,14 +123,31 @@ public class NativeLibraryLoader {
                 try {
                     // load library from user defined absolute path provided via pi4j.library.path}
                     System.load(path);
-                }
-                catch (Exception ex){
-                    //throw this error
-                    throw new UnsatisfiedLinkError("Pi4J was unable load the native library [" +
-                        libName + "] from the user defined library path.  The system property 'pi4j.library.path' is defined as [" +
-                        libpath + "]. Please make sure the defined the 'pi4j.library.path' " +
-                        "system property contains the correct absolute library path." +
-                        "; UNDERLYING EXCEPTION: [" + ex.getClass().getName() + "]=" + ex.getMessage());
+                } catch (UnsatisfiedLinkError ex) {
+                    String exceptMessage;
+                    // no guarantee the except pertains to ELF miss-match so check MSG content
+                    if (ex.getMessage().contains("wrong ELF class")) {
+                        exceptMessage = "Pi4J was unable to link the native library [" +
+                            path + "] embedded  inside this JAR [" +
+                            NativeLibraryLoader.class.getProtectionDomain().getCodeSource().getLocation().getPath() +
+                            "].  The exception indicates a mismatch of architecture. armhf/ELFCLASS32 aarch64/ELFCLASS64 \n" +
+                            " UNDERLYING EXCEPTION: [" + ex.getClass().getName() + "]=" + ex.getMessage();
+                    } else {
+                        exceptMessage = "Pi4J was unable to extract and load the native library [" +
+                            path + "] from the embedded resources inside this JAR [" +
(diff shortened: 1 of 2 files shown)


taartspi:
Frank commented on a post where the user was told they had a mixture of 32 and 64 bit shared objects.  He asked if we could provide some explanation in the exception we throw. Since the condition does not throw a unique excep there is no clean way to know this happened. So I check the error message to see if this binary mis-match occurred.  I updated the two normal paths in both the linuxfs and pigpio loader.  Could not test these error paths in the linuxfs loader as the test case I use does run, but  linuxfs does not read the gpio.

libraries/pi4j-library-linuxfs/src/main/java/com/pi4j/library/linuxfs/util/NativeLibraryLoader.java
```diff
@@ -124,13 +124,23 @@ else if (libpath.equalsIgnoreCase("local")) {
...
-                        "; UNDERLYING EXCEPTION: [" + ex.getClass().getName() + "]=" + ex.getMessage());
+                catch (Exception  | UnsatisfiedLinkError ex){
+                    String exceptMessage;
+                    if (ex.getMessage().contains("wrong ELF class")) {
+                        exceptMessage = "Pi4J was unable to link the native library [" +
+                            path + "] embedded  inside this JAR [" +
+                            NativeLibraryLoader.class.getProtectionDomain().getCodeSource().getLocation().getPath() +
+                            "].  The exception indicates a mismatch of architecture. armhf/ELFCLASS32 aarch64/ELFCLASS64 \n" +
+                            " UNDERLYING EXCEPTION: [" + ex.getClass().getName() + "]=" + ex.getMessage();
+                    }else{
```
eitch, line 135: can you fix the formatting?

libraries/pi4j-library-linuxfs/src/main/java/com/pi4j/library/linuxfs/util/NativeLibraryLoader.java
```diff
@@ -124,13 +124,23 @@ else if (libpath.equalsIgnoreCase("local")) {
...
                     System.load(path);
                 }
-                catch (Exception ex){
-                    //throw this error
-                    throw new UnsatisfiedLinkError("Pi4J was unable load the native library [" +
-                        libName + "] from the user defined library path.  The system property 'pi4j.library.path' is defined as [" +
-                        libpath + "]. Please make sure the defined the 'pi4j.library.path' " +
-                        "system property contains the correct absolute library path." +
-                        "; UNDERLYING EXCEPTION: [" + ex.getClass().getName() + "]=" + ex.getMessage());
+                catch (Exception  | UnsatisfiedLinkError ex){
```
eitch, line 127: can you put the UnsatisfiedLinkError in its own block?

libraries/pi4j-library-linuxfs/src/main/java/com/pi4j/library/linuxfs/util/NativeLibraryLoader.java
```diff
@@ -124,13 +124,23 @@ else if (libpath.equalsIgnoreCase("local")) {
                     // load library from user defined absolute path provided via pi4j.library.path}
                     System.load(path);
                 }
-                catch (Exception ex){
```
eitch, line 127: here the formatting is also not ok

eitch (commented):
I like your changes, thanks a lot. Can you please format your code, and put the UnsatisfiedLinkError in their own blocks? Since you're using an if/else, you might as well put it in their own block

taartspi:
Did a second push/pull.   Made sure my workspace had the .editconfig changes as shown on the web.  I only reformatted the area I changed.
Broke the catches into individuals.   As I don't know what conditions might throw UnsatisfiedLinkError I did leave the search for the wrong ELF class. Believer that bad messages are worse than none, so if not certain this is and ELF width error the more generic message is rethrown.

eitch (approved):

eitch:
Ok, merged it now

PR #248, opened 2022-10-16
track weather the user set the mode or bus config values. If they did…

taartspi:
Track weather the user set the mode and/or bus config values. If they supplied mode update flag  with the supplied value, if they supplied bus  update the flag with the supplied value. If no user supplied  mode skip flag update, no user supplied bus skip flag update.

 As the default value of mode and bus are both zero and the flags if not user supplied init value is zero, this will not alter the previous result when no user mode or bus was user supplied.

I reviewed the other providers and SPI is the only one that modifies the user provided config values.  I saw two ways to changes this, make the mode and bus required, or decided upon when to update the flags based on whether the user supplied mode and/or bus.  I choose the update based on user provided config data.

If what I did seems too difficult to explain in a point release README it would be few changes to make the mode and bus required,  and that is an easy explanation in the README


eitch (approved):
Looks fine to me

PR #251, opened 2022-11-02
Implement blink method in DigitalOutputBase

GeVanCo:
### blink() method

I needed the so-far unimplemented `blink()` method of the `DigitalOutputBase` class.  So I implemented it based on the following interpretation from my side:

1. `delay`: the time the output is high or low
2. `duration`: the amount of time the output has to switch from low to high or high to low

Example: 
- suppose you connect an LED to an output of the Raspberry Pi
- suppose you pass the following values for `delay`, `duration`, `unit` and `state`:

1. `delay`: 1
2. `duration`: 6
3. `unit`: `TimeUnit.SECONDS`
4. `state`: `DigitalState.HIGH` (implicit behaviour when you call `.blink(1, 6, TimeUnit.SECONDS);`)

Then the LED would have the following behaviour:

ON - 1sec - OFF - 1sec - ON - 1sec - OFF - 1sec - ON - 1sec - OFF - 1sec

So, the LED would visually be "ON" for 3 times, NOT 6 (as the `duration` parameter would suggest)!

I also added the above mechanism explanation in the header of the blink() method with a kind of graphical representation of the flow.  See the commit content.

### Solving an issue

Next to the implementation of the `blink()` method, I also discovered there was a mistake when comparing the `this.state` with the incoming `state` parameter for the method `state()`.  The `DigitalOutputBase` object itself was compared with the incoming state instead of the `state` variable itself.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java
@@ -74,7 +74,7 @@ public abstract class DigitalOutputBase extends DigitalBase<DigitalOutput, Digit
     @Override
     public DigitalOutput state(DigitalState state) throws IOException {
 
-        if(!this.equals(state)){
+        if(!this.state.equals(state)){
             this.state = state;
             this.dispatch(new DigitalStateChangeEvent<DigitalOutputBase>(this, this.state));
         }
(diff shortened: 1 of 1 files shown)


FDelporte:
Thanks @GeVanCo for your contribution! 
I need to further check the thread-use.

GeVanCo:
Oops... I see there's a calculation mistake for the hours.  Should be `(delay * 1000 * 60 * 60)` instead of `(delay * 1000 * 60 * 24)`.
I will push a new commit immediately.

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java
```diff
@@ -128,10 +128,100 @@ public Future<?> pulseAsync(int interval, TimeUnit unit, DigitalState state, Cal
...
+            case SECONDS:
+                millis = (delay * 1000);
+                break;
+
+            case MINUTES:
+                millis = (delay * 1000 * 60);
+                break;
+
+            case HOURS:
+                millis = (delay * 1000 * 60 * 60);
```
eitch, line 191: Can you please use the method: TimeUnit.toMillis()

This way you don't need to worry if you properly implemented your millisecond calculation, as the JDK already does this for you.

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java
```diff
@@ -128,10 +128,100 @@ public Future<?> pulseAsync(int interval, TimeUnit unit, DigitalState state, Cal
...
+            default:
+                throw new IllegalArgumentException("TimeUnit provided is not supported.");
+        }
+
+        this.state(state);
+
+        for (int i = 0; i < duration; i++) {
+            // block the current thread for the pulse duration
+            try {
+                Thread.sleep(millis);
```
eitch, line 203: Performing a Thread.sleep works, but it makes this method blocking. We should rather use a ScheduledExecutor which can perform this task for us and makes this method non-blocking, which i am sure is mostly desired.

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java
```diff
@@ -128,10 +128,100 @@ public Future<?> pulseAsync(int interval, TimeUnit unit, DigitalState state, Cal
...
+                throw new RuntimeException("Pulse blocking thread interrupted. Exception message: [" + e.getMessage() + "].");
+            }
+
+
+            // toggle the pulse state
+            this.state(DigitalState.getInverseState(this.state));
+        }
+
+        // invoke callback if one was defined
+        if (callback != null) {
```
eitch, line 215: If this method is blocking, then this callback isn't really useful, but if it would be non-blocking, then this would be excellent.

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java
```diff
@@ -74,7 +74,7 @@ public DigitalOutput initialize(Context context) throws InitializeException {
     @Override
     public DigitalOutput state(DigitalState state) throws IOException {
 
-        if(!this.equals(state)){
+        if(!this.state.equals(state)){
```
eitch, line 77: nice catch!

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java
eitch: ok, i just saw that we have a pulse() and a pulseAsync, so we can make this method blocking, but then we need to implement the async method as well...

eitch (changes requested):

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java
GeVanCo: This was a good suggestion! Implemented the TimeUnit.toMillis() in all the changes I've done (pulse(), pulseAsync(), blink(), blinkAsync()).

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java
GeVanCo: Indeed, I wrote you a personal e-mail telling you there was a blink() and blinkAsync(), just like there is a pulse() and pulseAsync() method already available.  
I've kept both blink() and pulse() methods as blocking methods and implemented the blinkAsync() and pulseAsync() methods with the ScheduledExecutorService.  Used a scheduleAtFixedRate() call for the blinkAsync() and a schedule() for the pulseAsync() method.

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputBase.java
GeVanCo: Well, maybe, maybe not.  In any case the caller will be informed about the end of the blocking phase and can still do whatever (s)he wants in the callback method.

GeVanCo:
Updated my previous commits with following changes:

- Took into account the suggestions/remarks from Robert for the blink() method
- Added/implemented the blinkAsync() method
- Cleaned up the pulse() method
- Added/implemented the pulseAsync() method

PR #254, opened 2022-11-28
Improve the readability of PiGpioNativeImpl#gpioInitialise and make it more robust

bwaldvogel:
While trying to read and understand `PiGpioNativeImpl#gpioInitialise` I struggled quite a bit to understand what the parameter value `3` of the `gpioCfgInterfaces` call actually means. So in my [first commit](4e24214a71448e6e61d808b400e895a497509b5c) I’m suggesting to dissolve the magic value `3` by using new constants that I’m adding to `PiGpioConst`. I’ve taken the constants from https://github.com/joan2937/pigpio/blob/master/pigpio.h.

In my [second commit](e953d99aad1ef403a9ba2af67fa8d8b1c39dcaf9) I’m suggesting to also validate the return value of the call to `gpioCfgInterfaces`. Currently Pi4J wouldn’t notice when pigpio returns an error in the `gpioCfgInterfaces` call.

Let me know what you think about my suggested improvements.

In a potential follow-up improvement, users could also be allowed to add the `PI_DISABLE_ALERT` flag¹. This was actually the triggering reason for my research since I saw about 5-7% CPU load on idle on my Raspberry Pi 3. Apparently it’s a known behavior of pigpio which is discussed in https://github.com/joan2937/pigpio/issues/29. Users who don’t need to wait for input events might want to add the `PI_DISABLE_ALERT` flag to shave off some unnecessary CPU load.

¹ _It is the same as adding the `-m` parameter to [`pigpiod`](https://abyz.me.uk/rpi/pigpio/pigpiod.html)_

### DIFF
--- a/libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/PiGpioConst.java
+++ b/libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/PiGpioConst.java
@@ -414,4 +414,12 @@ public interface PiGpioConst {
     int SIGTERM = 15;
     int SIGSTOP = 19;
     int SIGTSTP = 20;
+
+    // ----------------------------------
+    // IF FLAGS
+    // ----------------------------------
+    int PI_DISABLE_FIFO_IF = 1;
+    int PI_DISABLE_SOCK_IF = 2;
+    int PI_LOCALHOST_SOCK_IF = 4;
+    int PI_DISABLE_ALERT = 8;
 }
--- a/libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/PiGpioConst.java
+++ b/libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/PiGpioConst.java
@@ -418,8 +418,8 @@ public interface PiGpioConst {
     // ----------------------------------
     // IF FLAGS
     // ----------------------------------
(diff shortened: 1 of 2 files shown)


FDelporte:
Thanks, that's indeed a nice readability improvement as "magic numbers" are considered bad practice...
may replace IF with full description to make it even more clear?

personally I prefer the common part to be the start, but maybe that's just my feeling...

```
int PI_IF_DISABLE_FIFO = 1;
int PI_IF_DISABLE_SOCK = 2;
int PI_IF_LOCALHOST_SOCK = 4;
int PI_DISABLE_ALERT = 8;
```

bwaldvogel:
I’m happy to change the constants to be consistent to your naming conventions. I.e. I don’t have any (strong) opinion here.
However, wouldn’t it make sense to use the same names as `pigpio` does?

FDelporte:
@eitch what do you think?

bwaldvogel:
You’re welcome. I’ve renamed the constants in 2c120a10af37e3c483e907cbba17c786dd17eb25 as you suggested. What do you think about the inline comments that I’ve added which contain the names of the constants in pigpio?

Please also let me know if you would prefer that I squash 2c120a10af37e3c483e907cbba17c786dd17eb25 into 4e24214a71448e6e61d808b400e895a497509b5c before you merge this PR.

FDelporte (approved):

PR #262, opened 2022-12-26
Update LinuxFsI2C.java

harlanhu:
use offset.


FDelporte:
Thanks for your commit, @harlanhu. Did you see if this is something we missed in a unit test or don't have a test for?

eitch:
Hi @harlanhu, nice catch! I really did forget to use the offset.

Issue #303, opened 2023-10-18
Data read over I2C becomes out of sync over a slower wireless network


IAmNickNack:
I have come across a similar issue to #16 when using I2C via the GPIO socket implementation over a slow wireless network. I did not observe the problem when running the same code locally on the Pi via the loopback interface.

I observed that any payload was intermittently ignored when requesting multiple bytes over I2C and would be delivered as preamble prior to the next requested packet. 

For example, the request for 25 bytes, made over a slow network returns zero bytes:

```
c.p.l.pigpio.impl.PiGpioSocketImpl       : [I2C::READ] -> [0]; Register [18]; I2C Block [25 bytes]; offset=0
c.p.l.pigpio.impl.PiGpioSocketBase       : [TX] -> CMD=I2CRI(67); P1=0; P2=18; P3=4; PAYLOAD=[0x19 00 00 00]
c.p.l.pigpio.impl.PiGpioSocketBase       : [RX] <- CMD=I2CRI(67); P1=0; P2=18; P3=25; PAYLOAD=[0x]
c.p.l.pigpio.impl.PiGpioSocketImpl       : [I2C::READ] <- HANDLE=0; SUCCESS=true; RESULT=25
```

... the subsequent request then returns the payload:

```
c.p.l.pigpio.impl.PiGpioSocketImpl       : [I2C::WRITE] -> [0]; Register [16]; I2C Block [2 bytes]; offset=0
c.p.l.pigpio.impl.PiGpioSocketBase       : [TX] -> CMD=I2CWI(68); P1=0; P2=16; P3=2; PAYLOAD=[0x00 00]
c.p.l.pigpio.impl.PiGpioSocketBase       : [RX] <- CMD=I2CRI(67); P1=0; P2=18; P3=25; PAYLOAD=[0x00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00]
c.p.l.pigpio.impl.PiGpioSocketImpl       : [I2C::WRITE] <- HANDLE=0; SUCCESS=true; RESULT=25
```

On the other hand, the same request made via sockets when using the loopback interface returns the data as expected:

```
c.p.l.pigpio.impl.PiGpioSocketImpl       : [I2C::READ] -> [0]; Register [18]; I2C Block [25 bytes]; offset=0
[...]

PR #304, opened 2023-10-18
Read expected payload size of GPIO packet

IAmNickNack:
This PR addresses issue #303 and could possibly be expanded to cover SPI and serial data too.

### DIFF
--- a/libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/PiGpioPacket.java
+++ b/libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/PiGpioPacket.java
@@ -350,23 +350,17 @@ public class PiGpioPacket {
      * @throws IOException if an error occurs accessing {@code stream}.
      */
     public static PiGpioPacket decode(InputStream stream) throws IOException {
-        return PiGpioPacket.decode(stream.readNBytes(stream.available()));
-    }
-
-    /**
-     * <p>decode.</p>
-     *
-     * @param data an array of {@link byte} objects.
-     * @return a {@link com.pi4j.library.pigpio.PiGpioPacket} object.
-     */
-    public static PiGpioPacket decode(byte[] data) {
-        ByteBuffer rx = ByteBuffer.wrap(data);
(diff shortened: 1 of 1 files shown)


eitch:
Hi @IAmNickNack 
The code looks quite ok. Did you test it on a normal I2C connection as well?

IAmNickNack:
Could you clarify `normal I2C`?

If you mean I2C via non-socket GPIO, then no. Even prior to making this change I have been unable to get non-socket GPIO working on my local setup. Startup fails on my v4 device with `PIGPIO ERROR: PI_INIT_FAILED`. I have resorted to either GPIO over sockets or Linuxfs providers.

If this is what you mean. I can persevere.

eitch:
Hi @IAmNickNack the PI_INIT_FAILED error you get is most probably because you are not running it as root. Can you try again with root?

IAmNickNack:
> Hi @IAmNickNack the PI_INIT_FAILED error you get is most probably because you are not running it as root. Can you try again with root?

Running as root appears to make no difference. I will have to try a simpler test case than the application I'm currently working on to see if I can find out what's going on.

IAmNickNack:
> Hi @IAmNickNack the PI_INIT_FAILED error you get is most probably because you are not running it as root. Can you try again with root?

A-ha! This could be useful information to others, as I've seen this error mentioned in a number of posts:

Non-socket GPIO doesn't work when `pigpiod` is running! In my case it runs as a service to support development from a laptop. `pigpiod` presumably has some lock on `/dev/gpiomem` (?) 🤷 

I did still need to be running as root for this to work though.

So, in response to your original question @eitch, yes. I have tested this code change with a normal, non-socket i2c connection and have satisfied myself that it is good. 

Please feel free to suggest a more rigorous and definitive testing strategy to me if you like thought. 

Thanks

eitch:
Hi @IAmNickNack thanks for your work! I have merged your PR. Happy to see it works!

IAmNickNack:
Thanks @eitch. Hopefully somebody with an active serial or SPI project can add appropriate support for the equivalent operations to the `switch` as necessary.

Issue #305, opened 2023-10-18
Use Socket#setSoTimeout to timeout read requests of GPIO socket impl


IAmNickNack:

eitch:
I totally agree, your PR #306 is pending

eitch:
Your PR is merged

PR #306, opened 2023-10-18
Use Socket SO_TIMEOUT to wait for data

IAmNickNack:
This PR proposes an alternative implementation to wait for expected data when using the sockets GPIO provider as referred to in #305


eitch:
@IAmNickNack this is definitely better than the busy waiting. Thanks for the PR

PR #315, opened 2024-01-10
Extended Providers with a priority and have mock plugins marked as such

eitch:
### Extended Providers with a priority
This priority helps to determine which Provider should be loaded, when multiple providers with the same IOType are being loaded by different plugins.

This change enforces that a given IOType can only have one Provider loaded at runtime. Currently there is no known situation where this would be necessary, but could lead to problems if for instance two I2C providers are loaded at the same time, concurrently writing to the I2C bus.

Providers explicitly passed in the context config override the priority.

### Better handling of mock plugins
Plugins can now define if they are mocks, and these are not auto-detected anymore.

The default target for the pi4j library is the Raspberry Pi, and thus auto-detecting mocks on the Pi, which are only for tests is counterintuitive.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/runtime/impl/DefaultRuntime.java
+++ b/pi4j-core/src/main/java/com/pi4j/runtime/impl/DefaultRuntime.java
@@ -341,7 +341,7 @@ public class DefaultRuntime implements Runtime {
     }
 
     private void notifyInitListeners() {
-        // TODO
+        initializedEventManager.dispatch(new InitializedEvent(this.context));
     }
 
     @Override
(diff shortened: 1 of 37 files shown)


pi4j-core/src/main/java/com/pi4j/runtime/impl/DefaultRuntime.java
```diff
@@ -215,54 +217,68 @@ public Runtime initialize() throws InitializeException {
             plugins.clear();
 
             // container sets for providers and platforms to load
-            Set<Provider> providers = Collections.synchronizedSet(new HashSet<>());
-            Set<Platform> platforms = Collections.synchronizedSet(new HashSet<>());
+            Set<Platform> platforms = new HashSet<>();
```
FDelporte, line 220: out of curiosity, why isn't synchronizedSet not needed here anymore?

eitch: It was never necessary. Those are local variables, passed to the relevant classes, which then just copy/iterate the lists anyhow.

PR #333, opened 2024-03-10
linuxfs i2c compatibility with Pigpio and the i2c spec dictates these

taartspi:
i2c interface should use a restart between the write and read operation. --Combined format (see Figure 13). During a change of direction within a transfer, the START condition and the target address are both repeated, but with the R/W bit reversed.--
int i2cReadByteData(unsigned handle, unsigned i2cReg) int i2cReadWordData(unsigned handle, unsigned i2cReg)


eitch (approved):
Looks correct

Issue #335, opened 2024-03-15
java.io.IOException: Remote I/O error java.base/java.io.RandomAccessFile.writeBytes(Native Method)


taartspi:
Using linuxfs-i2c, dependent upon i2c  operations this exception can occur.   If the program initially uses  read or write, and later uses  readRegister or writeRegister there is no exception. However if the program  initially uses readRegister or writeRegister subsequent write or read may encounter this exception.                       As a temporary fix, when the program first create the i2c device, a read operation will be performed.     It is possible the attached i2c device always performs an  auto-increment in which case this additional read operation  will increment the device's internal pointer. As a program is not assured the state of any hardware, the program is already responsible to reset the chip or explicitly set such internal registers so this temporary change should show minimal effect.  When the proper fix is understood and implemented this 'extra' read will be removed.

eitch:
I currently see this fix as ok

taartspi:
Real fix found and implemented       Fix will build in snapHot 2.5.2

PR #336, opened 2024-03-15
Temporary fix for linuxfs-i2c  IOException. Also fix return code for …

taartspi:
…bytes read    readRegister.

See issue #335  for a complete description of the problem and temporary fix.

### DIFF
--- a/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/i2c/LinuxFsI2C.java
+++ b/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/i2c/LinuxFsI2C.java
@@ -316,7 +316,7 @@ public class LinuxFsI2C extends I2CBase implements I2C {
             buffer[i] = ioctlData.get(readBuffPosition +i );
 ;        }
 
-        return (ioctlData.limit());
+        return (readLength);
     }
 
 
@@ -330,8 +330,6 @@ public class LinuxFsI2C extends I2CBase implements I2C {
         reg[1] = (byte) (word & 0xff);
         reg[2] = (byte) ((word >> 8) & 0xff);
         byte buff[] = new byte[2];
-
-        int rCode = this.readRegister(reg, buff, 0, buff.length);
         word = (buff[1] << 8)  | buff[0];
         return word;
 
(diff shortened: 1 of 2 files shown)


plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/i2c/LinuxFsI2C.java
```diff
@@ -330,8 +330,6 @@ public int writeReadRegisterWord(int register, int word) {
         reg[1] = (byte) (word & 0xff);
         reg[2] = (byte) ((word >> 8) & 0xff);
         byte buff[] = new byte[2];
-
-        int rCode = this.readRegister(reg, buff, 0, buff.length);
```
eitch, line 334: Why did you remove this readRegister?

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/i2c/LinuxFsI2CProviderImpl.java
```diff
@@ -56,6 +56,8 @@ public synchronized I2C create(I2CConfig config) {
         LinuxFsI2CBus i2CBus = this.i2CBusMap.computeIfAbsent(config.getBus(), busNr -> new LinuxFsI2CBus(config));
         // create new I/O instance based on I/O config
         LinuxFsI2C i2C = new LinuxFsI2C(i2CBus, this, config);
+        // Workaround, needed if first LinuxFsI2C usage is ioctl (readRegister or writeRegister)
+        i2C.read();
```
eitch, line 60: Is there not an init of a buffer in this method, which then makes the ioctl work?

eitch: For instance i can see this:

        this.lastAddress = i2c.device();
        this.file.ioctl(I2CConstants.I2C_SLAVE, i2c.device() & 0xFF);

so the very first thing that happens, is that we set the lastAddress, then immediately perform an ioctl... So maybe this is missing?

eitch: ok, i do see that the readRegister() does this too...

eitch (approved):
I currently don't see a better solution, and as you wrote, a client is supposed to reset internal pointers anyhow. This code is approved

PR #337, opened 2024-03-20
Bumping versions

FDelporte:


eitch (approved):

eitch (changes requested):
@FDelporte 
Can you check why it doesn't build? Maybe we can't upgrade the ant-plugin just yet?

FDelporte:
> @FDelporte Can you check why it doesn't build? Maybe we can't upgrade the ant-plugin just yet?

Fixed, seems we can't bump this one:
`<ant-contrib.version>20020829</ant-contrib.version>`

PR #339, opened 2024-04-01
linuxfs-pwm provider fixes for Pi5 operation. Default now pwmchip2, w…

taartspi:
…as chip 0. After initial export of a channel, sleep quarter second so SSD card compllets directory and individual device file creation. Bug fix when user calls getActualFreq prior to setting PWM channel ON.

### DIFF
--- a/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/internal/LinuxPwm.java
+++ b/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/internal/LinuxPwm.java
@@ -44,7 +44,8 @@ public class LinuxPwm {
     public static String DEFAULT_SYSTEM_PATH = "/sys/class/pwm";
 
     /** Constant <code>DEFAULT_PWM_CHIP=0</code> */
-    public static int DEFAULT_PWM_CHIP = 0;
+    /** In Pi5 the chip is number 2 */
+    public static int DEFAULT_PWM_CHIP = 2;
 
     protected final String systemPath;
     protected final int chip;
--- a/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/pwm/LinuxFsPwm.java
+++ b/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/pwm/LinuxFsPwm.java
@@ -86,7 +86,7 @@ public class LinuxFsPwm extends PwmBase implements Pwm {
                 logger.trace("exporting PWM [" + this.config.address() + "]; " + pwm.getPwmPath());
                 pwm.export();
                 // Delay to allow the SSD to persist the new directory and device partitions
-                Thread.sleep(250);
+                Thread.sleep(70);
             } else{
                 logger.trace("PWM [" + this.config.address() + "] is already exported; " + pwm.getPwmPath());
             }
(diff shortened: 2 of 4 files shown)


plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/pwm/LinuxFsPwm.java
```diff
@@ -85,13 +85,18 @@ public Pwm initialize(Context context) throws InitializeException {
             if(!pwm.isExported()) {
                 logger.trace("exporting PWM [" + this.config.address() + "]; " + pwm.getPwmPath());
                 pwm.export();
+                // Delay to allow the SSD to persist the new directory and device partitions
+                Thread.sleep(250);
```
eitch: Is this really needed? Maybe the export method need to do this check? What fails without this sleep?

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/internal/LinuxPwm.java
```diff
@@ -44,7 +44,8 @@ public class LinuxPwm {
     public static String DEFAULT_SYSTEM_PATH = "/sys/class/pwm";
 
     /** Constant <code>DEFAULT_PWM_CHIP=0</code> */
-    public static int DEFAULT_PWM_CHIP = 0;
+    /** In Pi5 the chip is number 2 */
+    public static int DEFAULT_PWM_CHIP = 2;
```
eitch, line 48: Doesn't this mean, that it won't work on the Pi4 anymore? Shouldn't we be doing a check for the environment?

taartspi, line 48: test comment

eitch (changes requested):

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/pwm/LinuxFsPwm.java
taartspi: In the push last night the delay is 70 ms, a bit less pain.   The app does pwm.on()x,y);   If an export was not previously done, the code does an export and immediately call to set the period and calls to set the cycle.  Setting the period fails with an access violation.  In the export method we don't what will occur next and the delay fixes the exception.  My assumption is the SSD has not completed the creation of the pwmX directory and contained  files.   I used this code very little on my Pi4 long ago and it worked. When we have system aware code in use I can verify that the Pi4 does not have a problem with no delay.    Why 70 ms.  25 failed always, 30 worked nearly everytime.  So although 35 might have worked with my fast SSD I doubled that to 70 assuming users with lesser quality SSD will need a greater delay.

eitch: The export will create a virtual directory in the kernel's memory space. This should not have anything to do with the attached storage. But it is certainly possible, that it takes the kernel a few milliseconds longer to configure the hardware for this, so i guess we'll keep these 70ms.

taartspi: Ok   Stay 70.  I show my need to learn more on this.  Sometime maybe i will do more kernel learning to understand what is syncronous and also how the storage is really used

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/internal/LinuxPwm.java
taartspi: So yes. When franks Context update expose the machine type i can set the chip number correctly

eitch (approved):

Issue #340, opened 2024-04-04
Merge pi4j-board-info into core library


FDelporte:
To allow the easy detection of the Raspberry Pi board so plugins can define their priority and other settings, e.g. PWM address, based on the type of board.

Merge https://github.com/Pi4J/pi4j-board-info into this project

eitch:
@FDelporte I think this issue can be closed, right?

FDelporte:
correct @eitch , thanks for the cleanup ;-)

PR #341, opened 2024-04-04
#340 merging pi4j-board-info into pi4j-core

FDelporte:
To be able to use the data from the model in a REST API, this requires an extra dependency. Is this a problem?

```
com.fasterxml.jackson.core:jackson-annotations
```


eitch (changes requested):
We should perform the code to detect the board only once, and then store it in a singleton, or a class field on the BoardInfo class. It's quite heavy work to keep doing it, and it can't change at runtime anyhow.

eitch:
> To be able to use the data from the model in a REST API, this requires an extra dependency. Is this a problem?
> 
> ```
> com.fasterxml.jackson.core:jackson-annotations
> ```

Shouldn't we keep the JSON stuff out of core Pi4j? Let the REST API code generate the JSON code, then we don't need this dependency.

FDelporte:
> Shouldn't we keep the JSON stuff out of core Pi4j? 

@eitch I was expecting this ;-) is removed now...

FDelporte:
> We should perform the code to detect the board only once, and then store it in a singleton.

It's called from DefaultContext.java, so it's part of a singleton, correct?

https://github.com/Pi4J/pi4j-v2/pull/341/files#diff-34f18b9362bf258d05dac12071fa50bf49f05a9459c7875dbbb2044bdbe23962R107

eitch:
> > We should perform the code to detect the board only once, and then store it in a singleton.
> 
> It's called from DefaultContext.java, so it's part of a singleton, correct?
> 
> https://github.com/Pi4J/pi4j-v2/pull/341/files#diff-34f18b9362bf258d05dac12071fa50bf49f05a9459c7875dbbb2044bdbe23962R107

Sure, but maybe someone else would call BoardInfo.current(), as it is a static helper method.

FDelporte:
> Sure, but maybe someone else would call BoardInfo.current(), as it is a static helper method.

yep, in that case, it's up to the user to decide if they want to do that multiple times, right?

eitch:
> > Sure, but maybe someone else would call BoardInfo.current(), as it is a static helper method.
> 
> yep, in that case, it's up to the user to decide if they want to do that multiple times, right?

True, but from a library i expect it to avoid unnecessary computations. So multiple calls should cache this computation somehow.

Issue #244, opened 2022-09-22
Multiple SPI throws IOAlreadyExistsException


mvanassche:
When creating multiple spi, an exception IOAlreadyExistsException is thrown.

The config id is not transferred to the instance object.

Unlike other IO, SpiBase inherits from IOBase, which does not transfer the id/name to the instance (GpioBase and I2CBase do transfer).

I suppose there is simply 
```
        this.name = config.name();
        this.id = config.id();
        this.description = config.description();
```
missing in the constructor of SpiBase?

mvanassche:
```
diff --git a/pi4j-core/src/main/java/com/pi4j/io/spi/SpiBase.java b/pi4j-core/src/main/java/com/pi4j/io/spi/SpiBase.java
index 80f807e..751bc80 100644
--- a/pi4j-core/src/main/java/com/pi4j/io/spi/SpiBase.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/spi/SpiBase.java
@@ -48,6 +48,9 @@ public abstract class SpiBase extends IOBase<Spi, SpiConfig, SpiProvider> implem
      */
     public SpiBase(SpiProvider provider, SpiConfig config) {
         super(provider, config);
+        this.name = config.name();
+        this.id = config.id();
+        this.description = config.description();
     }
 
     /** {@inheritDoc} */
```

I tested with that diff with my example with two SPI devices on the same bus, different channel, and so far, it works.

FDelporte:
Thanks @mvanassche for reporting and proposing a fix!!!
I created a PR for this https://github.com/Pi4J/pi4j-v2/pull/245.

stefanhaustein:
It looks like this was fixed with #244 / #245

PR #344, opened 2024-04-14
fix(io): IOBase constructor do not update the fields id, name & description

### LINKED ISSUE #257: IOAlreadyExistsException when reopening Serial-connection
Hello.
I've got a problem when re-opening a serial connection using the class com.pi4j.io.serial.Serial.
Basically I want to write an interface class to receive data over an rs485-to-usb converter. And when the usb device gets disconnected and/or reconnected the software interface should close and reopen the serial connection. (Possibly the usb device could also change meanwhile - e.g. between ttyUSB0 and ttyUSB1, even when using the same usb slot.)
So the serial interface works fine, without problems, after application start. But whenever the interface gets closed and reopened the Context.create-function delivers the following IOAlreadyExistsException:

Stacktrace:
com.pi4j.io.exception.IOAlreadyExistsException: IO instance [com.pi4j.plugin.pigpio.provider.serial.PiGpioSerial] already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.

I closes the serial interface by using the Serial.close() function. I've also tried to use additionally the functions Serial.shutdown() and/or Context.getProvider().shutdown() and/or Context.getProvider().initialize(). But nothing worked for me. So the only way to change the serial device is now for me to restart the application.
I guess therefore that this is a bug and using the Serial.close() function doesn't unregister the used serial provider from the PI4J-Context. So I hope that someone here could verify this problem or has any idea what went wrong.
Alternatively maybe someone has an idea, how I could change the serial device on the fly - without having to close the serial connection.

My specifications:
- Raspberry 4B 4GB
- PI4J-Version: 2.2.1
- OS: Raspberry OS Lite (Debian Bullseye 32-bit)
- Pigpio-Version: 1.79-1+rpt1

One more note:
Using a raspberry-hat with a rs485-interface is unfortunately not really an option. Because some of the necessary I/O-Pins are still in use for other purposes.

fusetim:
The classes extending IOBase were already updating those fields (except SerialBase). SerialBase was the only class missing this change.

These changes allow the user to use multiple Serial interface at the same time. It was impossible before, as Pi4J would throw IOAlreadyExistsException.

This commit "moves" the fixes for the similar issues in the SPI interface. It also moves the similar code found in the dependency classes in the IOBase class.

Ref: #244, #257
Closes: #257

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/IOBase.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/IOBase.java
@@ -60,6 +60,9 @@ public abstract class IOBase<IO_TYPE extends IO, CONFIG_TYPE extends IOConfig, P
      */
     public IOBase(PROVIDER_TYPE provider, CONFIG_TYPE config){
         super();
+        this.id = config.id();
+        this.name = config.name();
+        this.description = config.description();
         this.provider = provider;
         this.config = config;
     }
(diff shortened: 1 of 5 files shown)


eitch (approved):
Thanks for that!

eitch:
You may change your PR to be ready for pulling

PR #351, opened 2024-04-18
[Major] I2C extensions, fixed and cleanup

eitch:
Added a I2C.execute(Callable<T>) method, allowing to atomically execute multiple I2C calls in a thread safe way

Added I2C.writeRead(byte[], byte[]) methods to atomically perform a write and then immediately afterwards a read on the I2C bus.

Fixed an issue where the LinuxFsI2CBus was closed when closing an I2C device - This was wrong, as another device might still be open on the same bus, and an operation on the underlying RandomAccessFile would lead to exceptions

Fixed the workaround requiring an I2C.read() on a newly created LinuxFsI2C device, if the first call was an ioctl - The device was not selected prior to the ioctl call.

Additional code cleanup, more to come in an additional commit

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/common/Action.java
+++ b/pi4j-core/src/main/java/com/pi4j/common/Action.java
@@ -1,6 +1,0 @@
-package com.pi4j.common;
-
-public interface Action {
-
-    void execute() throws Exception;
-}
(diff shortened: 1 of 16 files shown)


taartspi (approved):
Was the addition of the I2CBus classes to better represent the HW ?  
Also, was the move to callable to, as I understand, surface more details when there is an exception ?  
I have no comments on the code.

eitch:
@taartspi  the new I2CBus is to enable the execute() method on the I2C class. Thus allowing to perform a call with multiple read/writes atomically

eitch:
Rebased with develop after the other I2C changes

eitch:
@FDelporte should we merge?

PR #364, opened 2024-06-10
Don't load native code when not running on a recognized board

FDelporte:
Use board type to prevent errors while loading native libraries during development on non-Raspberry Pi.

@eitch can you please review if these are the right places to implement this? I have the feeling it should be somewhere "on a higher level", but not sure where...


eitch (approved):
Looks good! Every plugin does its own work, so i think this is fine. Should do it in the other plugins as well, where we don't have native libraries. pigpiod and LinuxFS. LinuxFS might work on a Linux platform, but not on Windows...

PR #365, opened 2024-06-18
Add runsOnRaspberryPi

FDelporte:
Extra boardinfo method to easily check if an app is running on a Raspberry Pi board.

Maybe we can still include this in V2.6.1?

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/boardinfo/util/BoardInfoHelper.java
+++ b/pi4j-core/src/main/java/com/pi4j/boardinfo/util/BoardInfoHelper.java
@@ -64,6 +64,10 @@ public class BoardInfoHelper {
         return instance.boardInfo.getBoardModel() == BoardModel.MODEL_5_B;
     }
 
+    public static boolean runsOnRaspberryPi() {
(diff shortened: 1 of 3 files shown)


taartspi (approved):
Looks good.    And i remember earlier conversations that linuxfs is expected to load ok on a foreign platform so this was not needed.

pi4j-core/src/main/java/com/pi4j/boardinfo/util/BoardInfoHelper.java
```diff
@@ -64,6 +64,10 @@ public static boolean usesRP1() {
         return instance.boardInfo.getBoardModel() == BoardModel.MODEL_5_B;
     }
 
+    public static boolean runsOnRaspberryPi() {
```
eitch: maybe we want to name it differently: runningOnRaspberryPi() ?

FDelporte: good point! will change

FDelporte: is done

eitch (approved):

Issue #369, opened 2024-07-26
WARN noise in the log about Ignoring providers on every startup


ylexus:
I have both gpiod (I need stable digital input and output) and linuxfs (I need i2c) plugins in my classpath. Both support digital i/o, so this results in
```
2024-07-26 23:26:35,286 WARN [main] c.p.r.i.DefaultRuntime Ignoring provider DIGITAL_INPUT LinuxFS Digital Input (GPIO) Provider with priority 100 as lower priority than GpioD Digital Input (GPIO) Provider which has priority 150
```
I am sure it's a valid use case for a user to want some providers from one plugin and some from another. So it's not a problem, but a BAU. 

In my application I have zero ERROR/WARN logging approach, so any ERROR or WARN is emailed to me. So on every startup I am receiving emails with the messages above.

Can the severity of this and similar messages be reduced to INFO?

ylexus:
Another example of what seems to should have been an INFO, if not DEBUG level message. Any reason it's a WARN?
```
2024-07-27 13:26:47,405 WARN [pi4j-shutdown] c.p.r.i.DefaultRuntimeRegistry Shutting down of IO DIN-25 took 24ms
```

FDelporte:
Correct remarks @ylexus. Log levels deserve a review. Do you happen to have some time to propose a merge request? Otherwise I will take a look in September after the holidays.

ylexus:
> Correct remarks @ylexus. Log levels deserve a review. Do you happen to have some time to propose a merge request? Otherwise I will take a look in September after the holidays.

Done, also done semi-automated removal of string concatenation in logging, except places where the thread safety of logging arguments is not certain.

eitch:
Fixed in next release

PR #370, opened 2024-07-27
fix unnecessary sting concatenation in logging statements; downgrade logging level to INFO in BAU places

ylexus:
#369

### DIFF
--- a/libraries/pi4j-library-gpiod/src/main/java/com/pi4j/library/gpiod/internal/GpioDContext.java
+++ b/libraries/pi4j-library-gpiod/src/main/java/com/pi4j/library/gpiod/internal/GpioDContext.java
@@ -70,7 +70,7 @@ public class GpioDContext implements Closeable {
             throw new IllegalStateException("Couldn't identify gpiochip!");
 
         this.gpioChip = found;
-        logger.info("Using chip " + this.gpioChip.getName() + " " + this.gpioChip.getLabel());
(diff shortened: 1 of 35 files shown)


FDelporte:
wow, thanks @ylexus, exactly the kind of review the logging needed! looks good af first sight, will dive in a bit deeper one of the next days

eitch (changes requested):
Thanks for the cleanup, but you changes some of the loggers from warning to info. Please undo that.

FDelporte:
> Thanks for the cleanup, but you changes some of the loggers from warning to info. Please undo that.

@eitch see ticket #369, some warnings at the start of the Pi4J context could, in my opinion, indeed be changed to info. Others we will need to review.

ylexus:
Yes, those changes to INFO was the whole point of this, the rest is just a side effect (automatic refactoring). Apologies, I should have mentioned the original issue is in the PR title and not in the first comment.

eitch:
You mixed two things in the same commit. Can you please change that, then i'll merge your request. Commits should clearly define what they change, which is not the case here

ylexus:
@eitch done

eitch (approved):
looks good

ylexus:
Thanks, how often do you do releases? Just to know when to expect this change to be released.

eitch:
As often as needed. We just did a 2.6.1. We might do a 2.6.2 soon.

PR #372, opened 2024-08-02
allowing configuring shutdown hook

eitch:
The shutdown hook is hooked into the JVM and is executed when the JVM stops.

This can be beneficial when the application is not managing pi4j's life cycle, but when an application wants to properly manage the life cycle, then this can be detrimental to the shutdown sequence of the application.

This change allows such an application to disable this feature.

The shutdown hook is still enabled by default, a future version might decide to disable it by default.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/context/ContextBuilder.java
+++ b/pi4j-core/src/main/java/com/pi4j/context/ContextBuilder.java
@@ -148,6 +148,34 @@ public interface ContextBuilder extends Builder<Context> {
     }
 
     /**
+     * <p>enableShutdownHook.</p>
+     *
+     * @return a {@link com.pi4j.context.ContextBuilder} object.
+     */
+    ContextBuilder enableShutdownHook();
+
+    /**
+     * <p>disableShutdownHook.</p>
+     *
+     * @return a {@link com.pi4j.context.ContextBuilder} object.
+     */
+    ContextBuilder disableShutdownHook();
+
+    /**
+     * <p>setShutdownHook.</p>
+     *
+     * @param enableShutdownHook a boolean.
+     *
+     * @return a {@link com.pi4j.context.ContextBuilder} object.
+     */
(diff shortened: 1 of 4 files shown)


FDelporte (approved):

eitch:
@FDelporte should we disable it by default, or leave it enabled as is currently the case?

FDelporte:
> @FDelporte should we disable it by default, or leave it enabled as is currently the case?

Not sure, I started doubting myself based on your remark in the ticket...

eitch:
Well in the end, it is quite simple for those who care. Change the following line:

    Context pi4j = Pi4J.newAutoContext();

to

    Context pi4j = Pi4J.newContextBuilder().autoDetect().disableShutdownHook().build();

or if it was default disabled, they can enable it again:

    Context pi4j = Pi4J.newContextBuilder().autoDetect().enableShutdownHook().build();

eitch:
I personally would prefer having it opt-in / disabled by default.

FDelporte:
That calls for some extra documentation ;-)
Will add with new release.

eitch:
Ok, i now disabled the hook by default. After the checks passed, I'll merge, and then think we can do a 2.6.2, or do you think this warrants a 2.7.0? @FDelporte

IAmNickNack:
I'm a little confused 😄

Which option most closely resembles the existing behaviour? 

It sounds like this decision could introduce a breaking change for existing users, but maybe I'm misreading this thread

eitch:
> I'm a little confused 😄
> 
> Which option most closely resembles the existing behaviour?
> 
> It sounds like this decision could introduce a breaking change for existing users, but maybe I'm misreading this thread

Yes, this is a breaking change. Existing behaviour can be reactivated with:

    Context pi4j = Pi4J.newContextBuilder().autoDetect().enableShutdownHook().build();

We will communicate this in the release notes, when we release.

PR #373, opened 2024-08-05
[Major] allow to load mocks on non-rpi systems

eitch:
This removes the

    Pi4J.newAutoContextAllowMocks()

function, as we now detect these systems using:

    BoardInfoHelper.runningOnRaspberryPi()

This is a breaking API change, but should satisfy people having to write special code to load the mock in tests, as this happens automatically now.

If a mock is to be loaded on the Raspberry Pi, then one can still do this using the following code:

    Pi4J.newContextBuilder().autoDetectMockPlugins().build();


FDelporte (approved):
Looks good!

PR #384, opened 2024-09-16
Utilize SPI Buffer size for large writes

mores:
Issue: https://github.com/Pi4J/pi4j-v2/issues/383


eitch (approved):
This looks good, thanks @mores

PR #391, opened 2024-10-05
Pi5 spi ioctl

taartspi:


taartspi:
@eitch @FDelporte   This is **NOT** for code review.  Doesn't have copyrights, formatting etc.... Not complete.     I pushed this code for conversation on how we should do the SPI for Pi5.   There also exist a zip in discussion #338 that contains an implementation offered by mpilone   My code has //to_do where I have open questions.  Also I implemented sufficient code to use ioctl and control a neopixel LED string.

PR #396, opened 2024-10-16
Add Checkstyle Plugin to Enforce Import Rules

dariuszzbyrad:
While analyzing the project, I found several unused and redundant import statements. To improve the code quality and ensure that best practices are followed going forward, I have added the Checkstyle plugin with rules to:

- Detect and remove unused imports.
- Prevent redundant imports (e.g., duplicate imports of the same class).
- Disallow illegal imports such as sun.* packages by default.

This configuration will help maintain clean and compliant code, and the build will fail if any violations are detected.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+    "https://checkstyle.org/dtds/configuration_1_3.dtd">
+
+<module name="Checker">
+    <module name="TreeWalker">
+        <module name="UnusedImports"/>
+        <module name="RedundantImport"/>
+        <module name="IllegalImport"/> <!-- defaults to sun.* packages -->
+    </module>
+</module>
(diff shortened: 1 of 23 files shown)


FDelporte:
Thanks @dariuszzbyrad! I'm all in favor of cleaner code, so this looks great. Asking for second opinion to Tom and/or Eitch so we can merge this.

taartspi (approved):
Thanks, this is a good idea to help keep the code clean.    You created a new _config_ directory with the context _checkstyle_.   Do you see more config file types  are possible and these separate directories are needed ?           Tom

dariuszzbyrad:
Yes, potentially at the config level, we could add tools like CodeNarc, PMD, SpotBugs, etc. Additionally, for Checkstyle specifically, we might include files like `suppressions.xml` and `suppressions-cpath.xml` to manage specific rule exceptions.

taartspi:
Thanks for the response and the PR.  Tom

eitch:
The issue looks fine, but we have to bear in mind, that our resources are limited. Adding new checks and then failing the build should not be a priority. I am always willing to look at PRs with new checks, where each new check is a new PR and the build doesn't fail with the new check.

If we can agree on this, then i find this a good change and am always for cleaner and better code.

eitch (approved):
This looks good

PR #426, opened 2024-11-27
WIP: Update Maven Plugin Versions and add Apache Maven Wrapper

dariuszzbyrad:
This PR introduces the Apache Maven Wrapper to the repository and updates the versions of Maven plugins to their latest stable releases. 

### Benefits:

-   **Consistency**: Ensures all developers and CI pipelines use the same Maven version.
-   **Convenience**: Simplifies setup for new developers, as Maven is automatically downloaded if unavailable locally.
-   **Portability**: Streamlines Maven usage across different operating systems (Windows, Linux, macOS).

### Next steps:
 - Update all instructions and documentation pages to reflect the use of the Maven Wrapper instead of installing Maven locally.


FDelporte:
Just an idea: wouldn't it be better to merge this into feature/#414 and keep this for the jump to Java 21?

dariuszzbyrad:
> Just an idea: wouldn't it be better to merge this into feature/#414 and keep this for the jump to Java 21?

Ok, will do

eitch:
To tell the truth i'm not particularly happy to commit binaries to the repo. I think the enforcement plugin is enough to force users to update their maven and have a local Maven properly configured.

dariuszzbyrad:
> To tell the truth i'm not particularly happy to commit binaries to the repo. I think the enforcement plugin is enough to force users to update their maven and have a local Maven properly configured.

Yes, definitely! The idea is to include just three files:

    The wrapper scripts (for Linux/MacOS and Windows)
    The configuration file (maven-wrapper.properties)

This approach ensures simplicity while avoiding the need to commit binaries to the repository.

dariuszzbyrad:
For better readability, PR has been created as new one (https://github.com/Pi4J/pi4j-v2/pull/427)

PR #428, opened 2024-11-29
DRAFT: #425: Adding CM5, but boardcodes are still missing...

### LINKED ISSUE #425: Extend boardinfo with Compute 5
https://www.youtube.com/watch?v=X4blR5Ua3S0

Need to find board code(s)...

FDelporte:
Closes #425

But board codes are still missing...


FDelporte:
Confirmed by @geerlingguy:

```
pi@cm5:~ $ cat /proc/cpuinfo | grep 'Revision' | awk '{print $3}'
c04180
That one is I think 4GB RAM mode
```

dariuszzbyrad (approved):

FDelporte:
Thanks for the approval @dariuszzbyrad, I just changed one more thing: converted code comment to remark in the enum so it will be visualized on api.pi4j.com

Issue #40, opened 2020-11-17
Unmapping/remapping pins


svnhub:
If I create an IO config for a pin with 
```
config = DigitalInput.newConfigBuilder(rpi4j).provider("pigpio-digital-input").id("pin-"+pinNumber).address(pinNumber)....
pin = rpi4j.create(config);
```
the name "pinX" is registered in the Registry and the pin works accordingly.

If I then want to disconnect the pin I call 
`pin.shutdown(rpi4j);`
which seems to forward the disconnect to pigpio but doesn't unregister the IO in the Registry.

Any attempt to then create a new config with the same id creates an exception:
```
com.pi4j.io.exception.IOAlreadyExistsException: IO instance [pin14] already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.
	at com.pi4j.provider.impl.ProviderProxyHandler.invoke(ProviderProxyHandler.java:99)
	at com.sun.proxy.$Proxy2.create(Unknown Source)
	at com.pi4j.context.Context.create(Context.java:309)
	at com.pi4j.internal.IOCreator.create(IOCreator.java:60)
	at com.pi4j.internal.IOCreator.create(IOCreator.java:101)
	at com.pi4j.internal.IOCreator.create(IOCreator.java:189)
        ...
```

The registry on the autoContext doesn't seem to have a way to remove things from it (it is a DefaultRegistry not RuntimeRegistry)

eitch:
This is a good point. I just ran into it myself.

beulenbilly:
Is there is a possibility to support you to solve the issue?

touchhome:
Does this thread alive?

FDelporte:
I see @eitch did a commit https://github.com/Pi4J/pi4j-v2/pull/198/files related to this topic.

eitch:
I merged a PR, and i think it should be fine now since v2.2.0

stefanhaustein:
I am still running into this issue with 2.7 and Context.shutdown(id) doesn't seem to be called from anywhere (I'd expect it to be called from (e.g.) DigitalBase.shutdown(context)?

Btw: Given that a context is required for creation (at least for DIO), isn't requiring a context for shutdown an (unnecessary?) opportunity for inconsistencies?

eitch:
Hi @stefanhaustein 

Please check this test: com.pi4j.test.registry.RegistryTest#testShutdownAndRecreate()

[...]

PR #438, opened 2025-01-20
(Currently failing) Test for shutdown and re-creation

stefanhaustein:
Illustrates that #40 isn't actually fixed.

I have tried to work around via manually calling config.shutdown(id). While this makes the registry problem disappear, I get a new error that seems to suggest that iobase.shutdown(config) doesn't actually release the corresponding resources properly.

### DIFF
--- a/pi4j-test/src/test/java/com/pi4j/test/registry/RegistryTest.java
+++ b/pi4j-test/src/test/java/com/pi4j/test/registry/RegistryTest.java
@@ -30,6 +30,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import com.pi4j.Pi4J;
 import com.pi4j.context.Context;
 import com.pi4j.exception.Pi4JException;
+import com.pi4j.io.gpio.digital.DigitalInput;
 import com.pi4j.registry.Registry;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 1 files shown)


eitch (approved):
Thanks for the test, we'll need to look into this

eitch:
Hi @stefanhaustein 

I finally managed to look at this issue, and the test is wrong. The issue is more a documentation issue: Don't shutdown the input directly, rather call Context.shutdown(id).

FDelporte:
@eitch can you please point out where this should be added in the documentation?

eitch:
@FDelporte The OP had an issue with shutdown, and he had tried to shutdown the IO by directly calling shutdown on it. This is not correct, as  one should pass the IO to be shutdown to the context. This will then perform the shutdown correctly. So i guess the documentation of the shutdown method on the IO should just say don't call this method directly, pass the ID to the context's shutdown method.

FDelporte:
@eitch this documentation change is OK?
https://github.com/Pi4J/pi4j.github.io/commit/870686719b9e02aa1d51d0709d808f7ec8b2801d

See https://www.pi4j.com/documentation/shutdown/

stefanhaustein:
Thanks for fixing the call. Please note that while #40 might be addressed, issue #440 still persists with the fixed call unfortunately, at least for GPIOD based digital input pins. 

Also, if the seemingly "obvious" call DigitalInput.shutdown() is not the correct call for properly shutting down a single input, but it can't be reduced in visibilty (I suppose), shouldn't it at least be named in a way that makes this more obvious, e.g. shutdownImpl or shutdownInternal?

PR #443, opened 2025-02-13
Allow Specifying GPIO Chip for GpioDContext

### LINKED ISSUE #439: Feature Request: Allow Specifying GPIO Chip for GpioDContext
## Current Behavior
Currently, the GpioDContext in pi4j-library-gpiod does not provide a way to explicitly specify the GPIO chip to use. It defaults to the first chip with a label containing "pinctrl" which might implicitly be /dev/gpiochip0 in many cases. However, this behavior is not guaranteed and may not be suitable for all use cases, especially when dealing with systems that have multiple GPIO chips or require using a specific chip other than the default.

## Expected Behavior
It should be possible to explicitly specify the GPIO chip to be used by GpioDContext, either by its name (e.g., "gpiochip2") or by its number (e.g., 2). If no chip is specified, GpioDContext should fall back to a default behavior, such as selecting the first chip with "pinctrl" in its label (the current behavior).

## Motivation
This enhancement would improve the flexibility and usability of GpioDContext in the following ways:
Support for Multiple GPIO Chips: Users working with systems having multiple GPIO chips can explicitly choose the desired chip.

## Proposed Solution
Introduce a setChip() method to GpioDContext that allows users to specify the desired GPIO chip. This method should accept either the chip name (String) or the chip number (int).

## Proposed Method Signatures:

```java
public synchronized void setChip(String chipName);
public synchronized void setChip(int chipNumber);
```

```java
public synchronized void setChip(String chipName) {
        if (this.gpioChip != null) {
            logger.warn("GpioD context already initialized with chip: {}.  Ignoring new chip request.", this.gpioChip.getName());
            return;
        }
        this.desiredChipName = chipName;
        logger.debug("GpioD chip name set to: {}", (chipName == null || chipName.isEmpty()) ? "[default]" : chipName);
    }
```

## Proposed Changes to initialize():

The initialize() method should be modified to:

```java
                if (this.desiredChipName != null && !this.desiredChipName.isEmpty()) {
                    if (chip.getName().equals(this.desiredChipName)) {
                        useThisChip = true;
                    }
                } else {
                    if (chip.getLabel().contains("pinctrl")) {
                        useThisChip = true;
                    }
                }
```

Haruka0522:
## Related issue
- close #439

### DIFF
--- a/libraries/pi4j-library-gpiod/src/main/java/com/pi4j/library/gpiod/internal/GpioDContext.java
+++ b/libraries/pi4j-library-gpiod/src/main/java/com/pi4j/library/gpiod/internal/GpioDContext.java
@@ -14,11 +14,7 @@ public class GpioDContext implements Closeable {
 
     private static final Logger logger = LoggerFactory.getLogger(GpioDContext.class);
 
-    private static final GpioDContext instance;
-
-    static {
-        instance = new GpioDContext();
-    }
+    private static final GpioDContext instance = new GpioDContext();
 
     public static GpioDContext getInstance() {
         return instance;
@@ -26,14 +22,47 @@ public class GpioDContext implements Closeable {
 
     private GpioChip gpioChip;
     private final Map<Integer, GpioLine> openLines;
-
     private final Set<Long> openLineEvents;
+    private String desiredChipName = null; // ユーザーが指定したチップ名を保持
 
     public GpioDContext() {
         this.openLines = new HashMap<>();
         this.openLineEvents = new HashSet<>();
     }
 
+    /**
+     * 使用する GPIO チップ名を設定します。
+     *
+     * @param chipName 使用する GPIO チップの名前 (例: "gpiochip0", "gpiochip2")。
+     *                 null または空文字列が指定された場合、デフォルトのチップが選択されます。
+     */
+    public synchronized void setChip(String chipName) {
+        if (this.gpioChip != null) {
+            logger.warn("GpioD context already initialized with chip: {}.  Ignoring new chip request.", this.gpioChip.getName());
+            return;
+        }
+        this.desiredChipName = chipName;
+        logger.debug("GpioD chip name set to: {}", (chipName == null || chipName.isEmpty()) ? "[default]" : chipName);
+    }
+
+    /**
+     * 使用する GPIO チップ番号を設定します。
+     *
+     * @param chipNumber 使用する GPIO チップの番号 (例: 0, 2)。
+     *                   負の値が指定された場合、デフォルトのチップが選択されます。
+     */
+    public synchronized void setChip(int chipNumber) {
+        if (chipNumber < 0) {
+            setChip(null);
+        } else {
+            setChip("gpiochip" + chipNumber);
+        }
+    }
+
+    /**
+     * GpioDContext を初期化します。
+     * setChip() でチップが指定されていない場合は、デフォルトのチップが選択されます。
+     */
     public synchronized void initialize() {
         if (!BoardInfoHelper.runningOnRaspberryPi()) {
             logger.warn("Can't initialize GpioD context, board model is unknown");
@@ -41,8 +70,10 @@ public class GpioDContext implements Closeable {
         }
 
         // already initialized
-        if (this.gpioChip != null)
+        if (this.gpioChip != null) {
+            logger.info("GpioD context already initialized with chip: {}", this.gpioChip.getName());
             return;
+        }
 
         long chipIterPtr = GpioD.chipIterNew();
         GpioChip found = null;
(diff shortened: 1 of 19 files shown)


FDelporte:
Hi @Haruka0522 thanks for contributing! Can you make the comments in English please? ;-)

Haruka0522:
Yes, sir.
I just noticed that, so I put it back in Draft. I will convert my comments to English soon.

Haruka0522:
I just updated my comment.

eitch:
From what i can see, this doesn't extend the configuration to set it from outside, does it?

Haruka0522:
Since GpioDContext is implemented as a singleton, users can modify the configuration by initializing the GpioDContext externally and calling the setChip method as needed.

eitch:
True, but the user should be using the standard pi4j configuration, which would mean one would apply it on the I2CConfigBuilder, and these fields would then be applied by DefaultI2CConfig

Haruka0522:
I understand your comment. It seems like a better way to implement as you commented, I will look into an implementation using ConfigBuilder that allows for proper chip selection.

Haruka0522:
The ContextBuilder is implemented so that the chip name can be specified with the setGpioChipName() method during initialization of the ContextBuilder.

plugins/pi4j-plugin-gpiod/src/main/java/com/pi4j/plugin/gpiod/provider/gpio/digital/GpioDDigitalInputProviderImpl.java
```diff
@@ -46,6 +46,11 @@ public int getPriority() {
     @Override
     public DigitalInputProvider initialize(Context context) throws InitializeException {
         DigitalInputProvider provider = super.initialize(context);
+        // Retrieve the chip name from the context properties
+        String chipName = context.properties().get("gpio.chip.name");
+        if (chipName != null) {
+            GpioDContext.getInstance().setChip(chipName);
+        }
```
eitch: I think it would make more sense to pass the properties to the initialize() method, in this way, we only need the property name once.

plugins/pi4j-plugin-gpiod/src/main/java/com/pi4j/plugin/gpiod/provider/gpio/digital/GpioDDigitalOutputProviderImpl.java
```diff
@@ -75,6 +75,11 @@ public int getPriority() {
     @Override
     public DigitalOutputProvider initialize(Context context) throws InitializeException {
         DigitalOutputProvider provider = super.initialize(context);
+        // Retrieve the chip name from the context properties
+        String chipName = context.properties().get("gpio.chip.name");
+        if (chipName != null) {
+            GpioDContext.getInstance().setChip(chipName);
+        }
```
eitch: and then we can skip this retrieval of the property.

eitch (changes requested):
Please see the changes in the comments

plugins/pi4j-plugin-gpiod/src/main/java/com/pi4j/plugin/gpiod/provider/gpio/digital/GpioDDigitalOutputProviderImpl.java
Haruka0522: Updated to pass properties directly.

libraries/pi4j-library-gpiod/src/main/java/com/pi4j/library/gpiod/internal/GpioDContext.java
```diff
@@ -26,37 +27,151 @@ public static GpioDContext getInstance() {
...
+            return;
+        }
+
+        // Set chip name from properties if available
+        String chipName = properties.get("gpio.chip.name");
+        if (chipName != null) {
+            setChip(chipName);
+        }
+
+        long chipIterPtr = GpioD.chipIterNew();
```
eitch, line 98: The code has now been duplicated, this is not a good solution. Please clean up the code.

Haruka0522: Duplicate code has been reorganized.
https://github.com/Pi4J/pi4j/pull/443/commits/ac77c63d08d7a883be05933c0a3b4c8288fab6e8

eitch (approved):
Looks good now. Thanks for the PR!

Haruka0522:
Thank you for your review and kind comments. It was a good learning experience for me.

eitch:
Happy to hear that. I was worried i was starting to annoy you with all the change requests. Let's see what @FDelporte has to say, then we can merge.

FDelporte (approved):
Looking good, thanks a lot

FDelporte:
@Haruka0522 can you propose some content for the documentation on how to use this new feature and in which circumstances?

Haruka0522:
Would a document structured like this be suitable?

**Overview**
On devices like industrial Raspberry Pis and Raspberry Pi 5 that are equipped with multiple GPIO chips, it is necessary to explicitly specify the target GPIO chip when using Pi4J to control GPIO. This prevents unintended access to other GPIO chips and enables accurate control.

**Checking GPIO Chips**
You can check the list of GPIO chips installed on the system using the gpiodetect command.
```
$ sudo gpiodetect
gpiochip0 [pinctrl-bcm2711] (58 lines)
gpiochip1 [raspberrypi-exp-gpio] (8 lines)
gpiochip2 [1-0027] (16 lines)
gpiochip3 [10-0020] (16 lines)
```
In the example above, four GPIO chips (gpiochip0, gpiochip1, gpiochip2, gpiochip3) are detected.

**Specifying GPIO Chips in Pi4J**
To control a specific GPIO chip with Pi4J, specify the GPIO chip name with the `.setGpioChipName()` method when creating a context using `Pi4J.newContextBuilder()` .
```
Context pi4j = Pi4J.newContextBuilder()
    .add(GpioDDigitalInputProvider.newInstance())
    .setGpioChipName("gpiochip2")
    .build();
```

taartspi:
Good explanation. Just need to specify this is gpiod  provider.  Unless
this is placed in a gpiod  specific area of the doc

eitch:
Thanks for the simple description.

FDelporte:
Thanks @Haruka0522, nice explanation! I guess I can add "By default, `gpiochip0` is used. You only need to specify the chip if you want to use another one."

FDelporte:
Is now documented here: https://www.pi4j.com/documentation/providers/gpiod/

FDelporte:
FYI, this is output on Pi4 with neW 3.0.0-SNAPSHOT, showing the output of this new code:

```
[main] INFO com.pi4j.library.gpiod.internal.GpioDContext - Using chip gpiochip0 pinctrl-bcm2711
[main] INFO com.pi4j.library.gpiod.internal.GpioDContext - GpioD context already initialized with chip: gpiochip0
```

PR #451, opened 2025-03-07
#308 Mark serial methods deprecated

FDelporte:


libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/internal/PIGPIO.java
```diff
@@ -55,12 +55,12 @@
...
- *      href="http://wiringpi.com/reference/shift-library/">http://wiringpi.com/reference/shift-library/</a>
  * @author Robert Savage (<a
- *         href="http://www.savagehomeautomation.com">http://www.savagehomeautomation.com</a>)
+ * href="http://www.savagehomeautomation.com">http://www.savagehomeautomation.com</a>)
  * @version $Id: $Id
+ * @see <a href="https://pi4j.com/">https://pi4j.com/</a>
+ * @see <a
+ * href="http://wiringpi.com/reference/shift-library/">http://wiringpi.com/reference/shift-library/</a>
  */
 public class PIGPIO {
```
taartspi, line 65: Should this one also be deprecated?

pi4j-core/src/main/java/com/pi4j/io/serial/SerialProviderBase.java
```diff
@@ -26,20 +26,22 @@
  * limitations under the License.
  * #L%
  */
+
```
taartspi, line 29: What about Serial. In this same package

libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/internal/PIGPIO.java
taartspi: That is the Serxxx functions

taartspi (approved):

libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/internal/PIGPIO.java
FDelporte: correct, done

pi4j-core/src/main/java/com/pi4j/io/serial/SerialProviderBase.java
FDelporte: correct, done

eitch (approved):
Looks good!

PR #452, opened 2025-03-17
 PWM linuxfs failed if the first interface call was off().  Fix is te…

taartspi:
…st if the accessed PWM address is enabled, is enabled normal code path, is not enabled do not attempt to disable.

The fix is line 164      The other edits are using the IDE to  Cleanup the file.     Looks like at creation my editor did not add a space between a "paren and brace"


FDelporte:
Thanks @taartspi !

PR #453, opened 2025-03-23
Make it possible to force a specific board model + improvements in setting usesRP1

FDelporte:


taartspi (approved):
Looks good.  I know there is a further change discussed in slack

PR #458, opened 2025-03-29
FFM API code base

DigitalSmile:
This PR is work in progress, will be adding new implementations step by step.

- [x] Overall design
- [x] Digital Input
- [x] Digital Output
- [x] i2c via smbus
- [x] i2c via ioctl
- [x] i2c via file write/read
- [x] spi
- [x] hardware pwm
- [x] uart

# Implementation details
FFM API provides new robust access to native code from JVM world. Moreover, all pi4j plugins relies on third party native libraries, which in the end calls linux kernel functions.
So, the typical flow was:
Java method (pi4j) -> JNA Java Code (jna) -> JNI Wrapper Call (pi4j) -> Native Library Call (gpiod, pigpio, etc) -> Kernel syscall
Now with the power of FFM API:
Java method (pi4j) -> Kernel syscall
This changes leads to better maintanability, less unsecure native calls, speed and better understanding of the code.

## Main package
New plugin consists of several packages:
1. `common` package with internal structures needed to FFM work
	- `file` - all needed code to access file descriptors with open/read/write/close/... methods
	- `gpio` - all needed structures to work with GPIO (DigitalInput/Output) interface
	- `i2c` - all needed structures to make I2C work (direct/ioctl/file). The only runtime dependency to work is `i2c-tools`
	- `ioctl` - syscall wrapper to work with linux kernel HAL
	- `poll` - syscall wrapper to work with poll/epoll messages
	- `serial` - all needed structures to work with Serial interface
	- `spi` - all needed structures to work with SPI bus
2. `providers` package with all user-facing implementations
	- `gpio` - contains DigitalInputFFM and DigitalOutputFFM implementations
	- `i2c` - contains I2CSMBus (using `i2c-tools` userspace library), I2CFile (using simple files) and I2CDirect (using `ioctl`)
	- `pwm` - contains PwmFFMHardware implementation
	- `serial` - contains SerialFFM implementation (beta version for now)
	- `spi` - contains SpiFFM implementation

## Tests
Since testing hardware is non-trivial tasks, I used an approach with different tests for different purposes.
Tests are organized in several packages:
1. `unit` -  classic unit testing with support of `Mockito` framework. Mocks are located at package `mocks`
2. `model` - tests, that covers model conversion from Java objects to FFM API structures
3. `integration` - tests that are using mock drivers on linux kernel side to simulate the behaviour of real hardware.
	- uses `i2c-stub`, `gpio-sim` and custom `spi-mock` kernel drivers (source code located in `native` folder)
	- the main sequence: setup drivers -> ensure all drivers are active -> make echo calls with pi4j -> test the return result
	- disabled for now, need to be moved to pi4j-harness project
4. `jmh` - classic JMH tests to measure speed of end-to-end calls with FFM/non-FFM plugins.

## JMH Test results
Pi4j has 3 providers that are used in common:
- `gpiod` provider with access to GPIO pins, uses `libgpiod` under the hood
- `pigpio` provider with GPIO, I2C, PWM, Serial and SPI, uses `libpigpio` under the hood with access through `/dev/mem`
- `linuxfs` provider with GPIO, I2C, PWM, Serial and SP, uses `sysfs` under the hood

All tests were run in VirtualBox environment using mock linux drivers:
- GPIO using `sim-gpio` driver
- I2C using `i2c-stub` driver
- SPI using custom self made echo driver
Setup scripts are located [here](https://github.com/Pi4J/pi4j/tree/ffm-api/plugins/pi4j-plugin-ffm/src/test/resources).

Because we are using mock drivers, we cannot test `pigpio` as it requires real hardware with available DMA addresses, so this library is out of scope of performance benchmark.
PWM is excluded from performance tests, because it is relying on filesystem in every provider available.
Serial is excluded, because FFM implementation is in beta and is out of scope with this PR.

### How to read results
The results should be analysed in terms of relative numbers, but not absolute. The less is average time - the best score it wins.

### GPIO (tested September, 20 2025)

- `OpenJDK Runtime Environment Corretto-25.0.0.36.2 (build 25+36-LTS)`
- `Linux ds-VirtualBox 6.14.0-29-generic #29~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Aug 14 16:52:50 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`
- `AMD Ryzen 7 5800X 8-Core Processor 29GiB System memory`
Results:
```
Benchmark                                                  Mode  Cnt     Score    Error  Units
GPIOPerformanceTest.testFFMInputRoundTrip                  avgt    5     0.173 ±  0.003  ms/op
GPIOPerformanceTest.testGpioDInputRoundTrip                avgt    5    10.537 ±  6.196  ms/op
GPIOPerformanceTest.testLinuxFsInputRoundTrip              avgt    5   111.781 ± 42.199  ms/op

Benchmark                                                  Mode  Cnt     Score    Error  Units
GPIOPerformanceTest.testFFMInputWithListenerRoundTrip      avgt    5     1.594 ±  5.830  ms/op
GPIOPerformanceTest.testGpioDInputWithListenerRoundTrip    avgt    5     9.015 ± 12.139  ms/op
GPIOPerformanceTest.testLinuxFsInputWithListenerRoundTrip  avgt    5   110.115 ± 18.796  ms/op

Benchmark                                                  Mode  Cnt     Score    Error  Units
GPIOPerformanceTest.testFFMOutputRoundTrip                 avgt    5     0.042 ±  0.001  ms/op
GPIOPerformanceTest.testGpioDOutputRoundTrip               avgt    5     0.110 ±  0.002  ms/op
GPIOPerformanceTest.testLinuxFsOutputRoundTrip             avgt    5   108.306 ± 12.490  ms/op

```

### SPI (tested September, 20 2025)
- `OpenJDK Runtime Environment Corretto-25.0.0.36.2 (build 25+36-LTS)`
- `Linux ds-VirtualBox 6.14.0-29-generic #29~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Aug 14 16:52:50 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`
- `AMD Ryzen 7 5800X 8-Core Processor 29GiB System memory`
Results:
```
Benchmark                                         Mode  Cnt  Score   Error  Units
SPIPerformanceTest.testFFMWriteReadRoundTrip      avgt    5  0.044 ± 0.006  ms/op
SPIPerformanceTest.testLinuxFsWriteReadRoundTrip  avgt    5  0.076 ± 0.001  ms/op
```
### I2C (tested September, 20 2025)
- `OpenJDK Runtime Environment Corretto-25.0.0.36.2 (build 25+36-LTS)`
- `Linux ds-VirtualBox 6.14.0-29-generic #29~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Aug 14 16:52:50 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`
- `AMD Ryzen 7 5800X 8-Core Processor 29GiB System memory`
Results:
```
Benchmark                                 Mode  Cnt  Score   Error  Units
I2CPerformanceTest.testFFMSMBusRoundTrip  avgt    5  0.038 ± 0.002  ms/op
```


plugins/pi4j-plugin-ffm/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+                <version>1.2.1</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>set-system-properties</goal>
+                        </goals>
+                        <configuration>
+                            <properties>
+                                <property>
+                                    <name>linux-headers</name>
```
eitch: Why do we need to set linux headers to a fixed version?

DigitalSmile:
For the history, benchmark for the simple roundtrip (create, get state, shutdown):
```
Result "com.pi4j.plugin.jmh.DigitalInputPerformanceTest.testCreateShutdownRoundTrip":
  1.652 ±(99.9%) 0.121 ms/op [Average]
  (min, avg, max) = (1.615, 1.652, 1.698), stdev = 0.031
  CI (99.9%): [1.531, 1.772] (assumes normal distribution)
```

eitch:
> For the history, benchmark for the simple roundtrip (create, get state, shutdown):
> 
> ```
> Result "com.pi4j.plugin.jmh.DigitalInputPerformanceTest.testCreateShutdownRoundTrip":
>   1.652 ±(99.9%) 0.121 ms/op [Average]
>   (min, avg, max) = (1.615, 1.652, 1.698), stdev = 0.031
>   CI (99.9%): [1.531, 1.772] (assumes normal distribution)
> ```

Perhaps you can add a benchmark for the existing provider?

DigitalSmile:
> > For the history, benchmark for the simple roundtrip (create, get state, shutdown):
> > ```
> > Result "com.pi4j.plugin.jmh.DigitalInputPerformanceTest.testCreateShutdownRoundTrip":
> >   1.652 ±(99.9%) 0.121 ms/op [Average]
> >   (min, avg, max) = (1.615, 1.652, 1.698), stdev = 0.031
> >   CI (99.9%): [1.531, 1.772] (assumes normal distribution)
> > ```
> 
> Perhaps you can add a benchmark for the existing provider?

That's for the the stage 2 :) I tried to quickly do it and failed, due to lack of build targets for desktop (amd64) of natives. I can try to run the benchmark on a real hardware, but since we are testing in comparison with other approaches, it is better to stay with gpio-sim.
Anyway will try to build locally later on.

DigitalSmile:
Okay, a bit of journey on benchmarks :)
### TLDR
For libgpiod (Java code -> JNA Code -> libpi4j-gpiod.so (JNI-wrapper) -> libgpiod.so (native library) -> GPIO kernel syscalls)
```
Result "com.pi4j.plugin.gpiod.jmh.DigitalInputPerformanceTest.testCreateShutdownRoundTrip":
  0.252 ±(99.9%) 0.015 ms/op [Average]
  (min, avg, max) = (0.248, 0.252, 0.257), stdev = 0.004
  CI (99.9%): [0.236, 0.267] (assumes normal distribution)
```
For FFM API (Java code -> GPIO kernel syscalls)
```
Result "com.pi4j.plugin.ffm.jmh.DigitalInputPerformanceTest.testCreateShutdownRoundTrip":
  0.167 ±(99.9%) 0.008 ms/op [Average]
  (min, avg, max) = (0.165, 0.167, 0.170), stdev = 0.002
  CI (99.9%): [0.159, 0.175] (assumes normal distribution)
```
### Bit of details
So first things first I built manually all gpiod natives (that was a lot more easier without maven) and pointed to the library folder in the code of `NativeLibraryLoader` class. Then I had to comment out lines in `GpioDContext` to prevent failing since I am not running Pi (btw, why there is such a limitation?).
My naive approach was to write the similar test like in https://github.com/Pi4J/pi4j/blob/e1f3079e19c0025271cb7bb947b6e05023e86038/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/jmh/DigitalInputPerformanceTest.java The problems begun immediately with SIGSEV from JVM, indicating that memory is corrupted.
`malloc(): unsorted double linked list corrupted`
I digged and found out, that threads created by `monitorLineEvents` are usually freeze and unable to shutdown properly. It seems the heap goes crazy in between native calls <-> monitoring thread <-> main java code. When I commented out the monitoring submission to ExecutorService all benchmarks went fine giving me above result. 

**_So for the sake of comparison I did the same thing with FFM code (commented out monitoring thread creation) and rerun benchmark._**

The results are as expected almost similar, with a slight faster with FFM. That happens, just because the roundtrip between native world and Java world is shorter.
Anyway, all above is jfyi, so I don't think we need to take any action items on that. I will continue with feature implementations. Next is i2c.

eitch:
Nice write up! Thanks a lot for that insight!

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/spi/FFMSpi.java
```diff
@@ -0,0 +1,135 @@
...
+        // get configured SPI mode
+        SpiMode mode = config.mode();
+
+        // the default value for 'flags' is zero
+        int flags = 0;
+
+        // if 'flags' were provided in the SPI config, then accept them
+        if(config().flags() != null){
+            flags = config().flags().intValue();
+        }
```
taartspi: in the linuxfs spi implementation we error out is the flags parm was used, This pertained to the pigpio provider and it continued use would put limits on the linuxfs code.

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/spi/FFMSpi.java
```diff
@@ -0,0 +1,135 @@
...
+
+        // only SPI BUS_0 and AUX SPI BUS_1 are supported by PiGPIO
+        if(bus.getBus() > 1){
+            throw new IOException("Unsupported BUS by PiGPIO SPI Provider: bus=" + bus.toString());
+        }
+
```
taartspi: also, ignoring the old pigpio limitations, the linuxfs spi will allow any bus and channel the user wants.  They are required to do the config.txt updates to the spi actually exists.

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/spi/FFMSpi.java
```diff
@@ -0,0 +1,135 @@
...
+            throw new IOException("Unsupported SPI channel (chip select) on SPI BUS_0 bus: address=" + config.address() );
+        }
+
+        // Comments on the PiGPIO web https://abyz.me.uk/rpi/pigpio/cif.html#spiOpen as follows:
+        // "Warning: modes 1 and 3 do not appear to work on the auxiliary SPI."
+        // SPI MODE_1 and MODE_3 are not supported on the AUX SPI BUS_1 by PiGPIO
+        if(bus == SpiBus.BUS_1 && (mode == SpiMode.MODE_1 || mode == SpiMode.MODE_3)) {
+            throw new IOException("Unsupported SPI mode on AUX SPI BUS_1: mode=" + mode.toString());
+        }
+
```
taartspi: broken record, this limitation does not exist in linuxfs

DigitalSmile:
@taartspi just for the record - SPI classes are just the boilerplate code now, I copy-pasted it for future reference. You can have a look on my implementation here (which I'll be using): https://github.com/DigitalSmile/gpio/blob/main/src/main/java/org/digitalsmile/gpio/spi/SPIBus.java

DigitalSmile:
Implemented i2c base classes, need further testing.

plugins/pi4j-plugin-ffm/src/test/resources/gpio-setup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+modprobe configfs
```
eitch, line 3: Why do we need this? Is this a new constraint to use FFM that we need to configure the kernel's GPIO before use? This might be quite an issue.

@FDelporte

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/I2CSMBus.java
```diff
@@ -0,0 +1,149 @@
...
+            } else {
+                return new byte[0];
+            }
+        });
+    }
+
+
+    @Override
+    public int read() {
+        i2CBus.selectDevice(config.device());
```
eitch: Is this thread safe? One thread might select a device, then another selects another device and then the first does the execute...

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/I2CDirect.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * Internal method.
+     * Writes the data array into the register address of device selected previously.
+     *
+     * @param data data array to be written
+     */
+    private int writeInternal(int register, byte[] data) {
+            i2CBus.selectDevice(config.device());
+            return i2CBus.execute(this, (i2cFileDescriptor) -> {
```
eitch: Thread safety here everywhere... The execute method should perform the device selection on the bus and keep a lock on the bus

eitch (commented):
Looking good, but there are still some open issues

plugins/pi4j-plugin-ffm/pom.xml
eitch: Should version the generated headers? What if a new kernel version requires a change, but we want to be backwards compatible with older kernels?

plugins/pi4j-plugin-ffm/src/test/resources/gpio-setup.sh
DigitalSmile: This is located in test/resources and is needed to run unit tests.

plugins/pi4j-plugin-ffm/pom.xml
DigitalSmile: Left from native-memory library, will remove

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/I2CSMBus.java
DigitalSmile: Indeed, not thread safety now :) Will add later on

plugins/pi4j-plugin-ffm/pom.xml
DigitalSmile: Well, it is already removed, you were looking outdated file...

plugins/pi4j-plugin-ffm/src/test/resources/gpio-setup.sh
DigitalSmile: So scripts are just preparing testing environment (setup and clean `gpio-sim` and `i2c-stub`)

eitch: Ok, good

DigitalSmile:
Created SPI and Hardware PWM

DigitalSmile:
### New performance tests (GPIO, I2C SMBus, SPI)
 ```
Benchmark                                           Mode  Cnt   Score    Error  Units
GPIOPerformanceTest.testInputRoundTrip              avgt    5   0.169 ±  0.005  ms/op
GPIOPerformanceTest.testInputWithListenerRoundTrip  avgt    5  11.766 ± 71.672  ms/op
GPIOPerformanceTest.testOutputRoundTrip             avgt    5   0.042 ±  0.003  ms/op
I2CPerformanceTest.testSMBusRoundTrip               avgt    5   0.035 ±  0.001  ms/op
SPIPerformanceTest.testWriteReadRoundTrip           avgt    5   0.037 ±  0.002  ms/op
```
### Details
What is measured: roundtrip throughput from building Pi4J context to shutdown, including simple operations (write, read or asking state).
Tests are run with mock linux drivers, which are simply echoing the result of write operations: `gpio-sim`, `i2c-stub`, self made `spi-mock` (https://github.com/Pi4J/pi4j/blob/ffm-api/plugins/pi4j-plugin-ffm/src/test/native/spi-mock.c).
### Analysis
GPIO Output, I2C and SPI are working as expected - the foot print and roundtrip time is minimal.
GPIO Input with state change listener consumes more time, because listener is implemented in a separate thread and there is thread management involved.
GPIO Plain Input has higher timings, need to investigate why it is times slower than GPIO Output.

If someone can write JMH tests for current implementations of gpiod/sysfs/pigpio that would be very nice.

File: plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/pwm/PwmFFMProviderImpl.java [hunk 8]
```diff
@@ -6,8 +6,8 @@
 public class PwmFFMProviderImpl extends PwmProviderBase implements PwmProvider {
 
     public PwmFFMProviderImpl() {
-        this.id = "PwmFFMProviderImpl";
-        this.name = "PwmFFMProviderImpl";
+        this.id = "ffm--pwm";
+        this.name = "FFM API Provider PWM";
```
taartspi, line 13: ffm-pwm.    There are two —

taartspi (changes requested):

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/pwm/PwmFFMProviderImpl.java
FDelporte: thanks, is fixed

taartspi:
Looks the merge is complete and ffm updates are ready for HW testing

taartspi (commented):
I have reviewed parts of the code not all. This delivery is a great deal of work. I will continue to review and test    I think we should merge and build this to make testing and any new issue discussions easier.

eitch:
So i'll update develop to 4.0.0-SNAPSHOT and then merge this branch.

Any objections? @FDelporte @DigitalSmile @taartspi

FDelporte:
@eitch the branch is already on `<version>4.0.0-SNAPSHOT</version>`. 
So yes, please, let's merge and speed up testing 🚀
There is only one direction: FORWARD 😉

eitch:
Done.

Issue #457, opened 2025-03-28
Why is SpiChipSelect limited to 0, 1, 2?


douglm:
Fairly new to this and trying to get widgetlords digital input boards working with a Raspberry Pi 5 and pi4j.

I hope I'm reading this right but cat /sys/kernel/debug/gpioio gives me
...
 gpio-575 (GPIO4               |spi0 CS2            ) out hi ACTIVE LOW
...
 gpio-578 (GPIO7               |spi0 CS1            ) out hi ACTIVE LOW
 gpio-579 (GPIO8               |spi0 CS0            ) out hi ACTIVE LOW
...
 gpio-588 (GPIO17              |spi0 CS3            ) out hi ACTIVE LOW
...
 gpio-593 (GPIO22              |spi0 CS4            ) out hi ACTIVE LOW 

and I see
crw-rw---- 1 root spi 153, 3 Mar 21 11:17 /dev/spidev0.0
crw-rw---- 1 root spi 153, 4 Mar 21 11:17 /dev/spidev0.1
crw-rw---- 1 root spi 153, 2 Mar 21 11:17 /dev/spidev0.2
crw-rw---- 1 root spi 153, 1 Mar 21 11:17 /dev/spidev0.3
crw-rw---- 1 root spi 153, 0 Mar 21 11:17 /dev/spidev0.4
crw-rw---- 1 root spi 153, 5 Mar 21 11:17 /dev/spidev10.0

The board uses GPIO17 or GPIO22 which I believe correspond to address 3 and 4. However - setting address 3 gives me an NPE as there is no SpiChipselect value for that address.

Is there any reason why this isn't allowed?

FDelporte:
Historically I guess...
Could it be that 3 and up was not available pre-RPI5?
In that case, we should change the code to allow a higher value.

douglm:
I rebuilt with up to CS_6 defined and my code now sees the state changes on the digital board. 

From my perspective (50 years programming - java since the beginning - RPI since a couple of weeks ago) the SpiChipSelect thing adds a layer of confusion. Shouldn't this - in LinuxFsSpi:

String spiDev = SPI_DEVICE_BASE + config().bus().getBus() + "." + config().getChipSelect().getChipSelect();

be 

String spiDev = SPI_DEVICE_BASE + config().bus().getBus() + "." + config().address();

SpiChipSelect introduces a limit on the addresses - my config was actually:

    final var config = Spi.newConfigBuilder(pi4j)
[...]

PR #461, opened 2025-04-09
SpiChipSelect located in the core code base contained code enforcing …

taartspi:
…a Pigpio limitation of chipselect 0 1 2 only allowe values.  Move this limit check in the Pigpio provider. Update SpiChipSelect to support values 0 thru 10. If a client required more than 11 GPIOs dedicated to chipSelects I think they would instead use a single MCP23017 i2c chip to control the CS, and save all the GPIOs.   

I think the on-going rewrite to abandon JNA and JNI should either create a brand new Spi Config builder or tear the existing one apart.  At present the user can use address, channel or chipSelect to assign the spi channel. If all three attribute are used at once, the last one one the build command is used. This just adds to the confusion of using the GPIO or true channel number introduced in linuxfs-spi.


taartspi:
And I validated using a bus or channel not configured on the machine creates an understandable error message

FDelporte:
Relates to issue #457

eitch (approved):
Looks good, thanks!

PR #462, opened 2025-04-09
Further improved SpiSelect

FDelporte:
@eitch you beat me in speed and already merged ;-)
I didn't like the many if checks, so made a few more changes in the merge request of @taartspi 

* Simplified if
* Added unit test
* Log on SPI open in LinuxFsSpi


eitch (approved):
You are of course right, Frank, this is much better!

PR #468, opened 2025-07-09
linuxfs-i2c  registerRead using offset parm to place read data, faile…

taartspi:
…d to use the 

My bug to ignore the user supplied offset

Intellij reformat would make many changes to the file.  I wanted to be certain of the correct format before letting the IDE rip.


taartspi:
#467  associated discussion

FDelporte:
@taartspi can this change be validated with a unit test?

taartspi:
No test. This method goes way back to pigpio and never tested this offset parm.  I will add the test

taartspi:
Looked at the tests we have.   The offset usage IS tested, but as those tests use the MockI2c, they test the Core provider    
common code but not the concrete providers like linuxfs.    
@FDelporte    This is complete, please merge if you agree.

PR #476, opened 2025-08-18
when the RP1 chip is found, PWM provider will use the RP1 architected…

taartspi:
… address of PWM0 to determine the correct pwmChipx to use.   This change will effect Pi5 and CM5 using updated kernel

Implemented code to allow multi digit pwmChip? values.  Not likely but reading the post from Pi engineers if pwmPIOs are created they may be numbered before the user PWM.

### DIFF
--- a/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/LinuxFsPlugin.java
+++ b/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/LinuxFsPlugin.java
@@ -29,24 +29,21 @@ package com.pi4j.plugin.linuxfs;
 
 
 import com.pi4j.boardinfo.util.BoardInfoHelper;
-import com.pi4j.boardinfo.util.command.CommandResult;
 import com.pi4j.extension.Plugin;
 import com.pi4j.extension.PluginService;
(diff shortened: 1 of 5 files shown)


plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/LinuxFsPlugin.java
```diff
@@ -128,6 +135,30 @@ public void initialize(PluginService service) {
         int pwmChip;
         if(BoardInfoHelper.usesRP1()) {
             pwmChip = LinuxPwm.DEFAULT_RP1_PWM_CHIP;
+            // init to original bookworm using pwmChip2, test if different
+            String command =  "ls -l " + pwmFileSystemPath  ;
+            CommandResult rslt = execute( command);
+
+            String[] paths =  rslt.getOutputMessage().split("\n");
+            int counter = 0;
```
FDelporte: can this be extracted to a seperate method + unit test? it's not really clear to me what this should read and return...

taartspi:
@FDelporte   Frank, I tested with the Optional usage, made a couple changes but your suggested means worked well, thanks.  As this is main line code I wanted to give you another shot at any comments before doing a merge. thanks Tom

Issue #478, opened 2025-08-24
Race condition in GpioDDigitalInput causes monitor thread to exit


mpilone:
Since updating to v3.0.2 I've run into an issue where a GPIOD input never fires a change event. I did some debugging and traced it down to [monitorLineEvents](https://github.com/Pi4J/pi4j/blob/7ae9f8a942b6414aedc21a1001445fe0b7d134b1/plugins/pi4j-plugin-gpiod/src/main/java/com/pi4j/plugin/gpiod/provider/gpio/digital/GpioDDigitalInput.java#L107). The while loop in that method uses `this.inputListener`, which is assigned a value as the result of submitting the `monitorLineEvents` to the executor. This causes a race condition where the monitor loop could start and attempt to use the inputListener before it has been assigned. 

Additionally, the `inputListener` isn't marked as volatile or protected in anyway which means the `monitorLineEvents` thread may not see the value assigned to inputListener even if it was assigned before the loop was executing.

With some additional logging, catching an exception on the `try` block of `monitorLineEvents`, I got:

```
[2025-08-23T23:49:12.001+0000] [INFO ] [ugin.gpiod.provider.gpio.digital.GpioDDigitalInput] [Pi4J.RUNTIME-2                     ]: monitorLineEvents: exception
java.lang.NullPointerException: Cannot invoke "java.util.concurrent.Future.isCancelled()" because "this.inputListener" is null
	at com.pi4j.plugin.gpiod.provider.gpio.digital.GpioDDigitalInput.monitorLineEvents(GpioDDigitalInput.java:113)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
```

It looks like these issues were introduced with commit [e92fabc](https://github.com/Pi4J/pi4j/commit/e92fabce6a3cacd09b753dcce18c153ecd632d8f) in an attempt to simplify the input implementation. In the earlier code, the monitorLineEvents loop would check if inputListener was null first which protected against this scenario. 

[...]

PR #479, opened 2025-08-25
[Fix] #478 Improve thread safety and shutdown handling in GpioDDigitalInput

eitch:
Fixed the subtle remaining race condition on initialization of the monitoring task


FDelporte:
@mpilone can you review please?

mpilone:
Looks good to me. I think the code could be simplified even more by checking `Thread.currentThread.isInterrupted()` in the monitor loop rather than a separate boolean and the shutdown code could wait on `inputListener.get()` rather than the separate countdown latch, but these changes should fix the immediate issue. Thanks for the quick fix.

mpilone (approved):

PR #483, opened 2025-09-05
Spi write large buffer

taartspi:
This implementation is same as found in pigpio    It also creates multiple spi writes  (ioctl call) I communicated with the pigpio author, them telling me the multi writes to get all the data into the spi device operated correctly.   

I have another version of this fix written, it uses multi transfers in an array that is passed on a single ioctl.  This would be a better solution.  I hit intermittent IOERRORs when using this code.  I will write a simple testcase to duplicate the error and post details to raspberryPi forum.    If I use this code with a single entry in the array it works 

I don't know how many user write more than 4k bytes.  And they can increase the buffer if needed.  Just this user needed 115200 bytes.

### DIFF
--- a/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
+++ b/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
@@ -42,6 +42,11 @@ import com.sun.jna.ptr.IntByReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.Scanner;
+
 import static com.pi4j.plugin.linuxfs.internal.LinuxLibC.*;
 
 /**
@@ -59,13 +64,14 @@ import static com.pi4j.plugin.linuxfs.internal.LinuxLibC.*;
  */
 public class LinuxFsSpi extends SpiBase implements Spi {
 
-    private static final Logger LOG = LoggerFactory.getLogger(LinuxFsSpi.class);
+    private static final Logger logger = LoggerFactory.getLogger(LinuxFsSpi.class);
 
     ///////////////////////////////////
     // From spidev.h
     private final static byte SPI_IOC_MAGIC = 'k';
     private final static byte SIZE_OF_BYTE = 1;
     private final static byte SIZE_OF_INT = 4;
+    private int SPI_BUFFSIZ = 4096;
 
     private static int SPI_IOC_MESSAGE(int n) {
         // Even though we will pass the structure to ioctl as a pointer, the command needs to know
@@ -141,6 +147,19 @@ public class LinuxFsSpi extends SpiBase implements Spi {
 
     public LinuxFsSpi(LinuxFsSpiProviderImpl provider, SpiConfig config) {
         super(provider, config);
+        SpiInitialise();
+    }
+
+    private void SpiInitialise() {
+        try {
+            Scanner scanner = new Scanner(new File("/sys/module/spidev/parameters/bufsiz"));
+            if (scanner.hasNextInt()) {
+                SPI_BUFFSIZ = scanner.nextInt();
+            }
+            logger.trace("[INITIALIZE] -> SPI_BUFFSIZ={}", SPI_BUFFSIZ);
+        } catch (Exception e) {
+            logger.error(e.getMessage(), e);
+        }
     }
 
     @Override
@@ -154,7 +173,7 @@ public class LinuxFsSpi extends SpiBase implements Spi {
         //    character special device, major number 153 with a dynamically chosen minor device number.
         //    This is the node that userspace programs will open, created by “udev” or “mdev”.
         String spiDev = SPI_DEVICE_BASE + config().bus().getBus() + "." + config().address();
-        LOG.info("Opening SPI bus {}, address {}", config().bus().getBus(), config().address());
+        logger.info("Opening SPI bus {}, address {}", config().bus().getBus(), config().address());
         fd = libc.open(spiDev, LinuxLibC.O_RDWR);
         if (fd < 0) {
             throw new RuntimeException("Failed to open SPI device " + spiDev);
@@ -259,7 +278,7 @@ public class LinuxFsSpi extends SpiBase implements Spi {
 
         int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(1), transfer);
         if (ret < 0) {
-            LOG.error("Could not write SPI message. ret {}, error: {}", ret, Native.getLastError());
+            logger.error("Could not write SPI message. ret {}, error: {}", ret, Native.getLastError());
             numberOfBytes = -1;
         }
         else {
@@ -296,7 +315,7 @@ public class LinuxFsSpi extends SpiBase implements Spi {
 
         int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(1), transfer);
         if (ret < 0) {
-            LOG.error("Could not write SPIIOC  message. ret {}, error: {}", ret, Native.getLastError());
+            logger.error("Could not write SPIIOC  message. ret {}, error: {}", ret, Native.getLastError());
             length = -1;
         }else{
             buf.read(0, read, offset, length);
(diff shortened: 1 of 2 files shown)


plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -312,29 +331,67 @@ public int write(byte b) {
...
+     * a single SPI write operation.
+     * So CE line low, write bytes, CE line high
+     *
+     * A write greater than 4096 bytes will be segmented to multiple writes
+     * each 4096  bytes in length, the last write is the MOD value.
+     * So, CE line low, write first 4096 bytes CE line high.  This
+     * pattern will repeat until the last bytes length MOD 4096 are written.
+     * This means multiple SPI transaction with you SPI device. If CE line
+     * toggling creates problems with your SPI device your application
+     * an use a vacant GPIO configured as an Output pin and your
```
FDelporte: "can use"

FDelporte (approved):

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -141,6 +147,19 @@ public static class ByValue extends spi_ioc_transfer implements Structure.ByValu
 
     public LinuxFsSpi(LinuxFsSpiProviderImpl provider, SpiConfig config) {
         super(provider, config);
+        SpiInitialise();
+    }
+
+    private void SpiInitialise() {
```
eitch: Can you make the method lower case?

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -312,29 +331,67 @@ public int write(byte b) {
...
-        transfer.delay_usecs = 0;
-        transfer.len = length;
-
-        int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(1), transfer);
-        if (ret < 0) {
-            LOG.error("Could not write SPI message. ret {}, error: {}", ret, Native.getLastError());
-            length = 0;
+        Objects.checkFromIndexSize(offset, length, data.length);
+
+        byte[] someData = Arrays.copyOfRange(data, offset, length);
```
eitch: Is this a defensive copy before running? Isn't this just a copy of the data to send?

eitch (changes requested):
I have a few change requests

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: done   
Since it is in the spi class shortened the name to initialize

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: done

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: was just a copy   not required so removed

taartspi:
believe I made the requested updates   and again.  I think we should consider NOT doing this in V4 FFM.    @eitch @FDelporte

eitch:
Why do you think we should not do that in v4 ffm?

eitch (approved):

taartspi:
Why not in V4.    And i am talking about a fix where a single ioctl sends all the data to the device. When/if i go ahead and get the multi transfers sent with a single ioctl working i can use an analyzer to measure the following.    I assume even the kernel will spend some measurable time moving thru the transfer array as one element is completed and the next element begun. If that delay in the bit stream cause any problem in the spi device operation it would be difficult for the user to debug.  Maybe being too dramatic on the possible problem but it would be a difficult debug.  @eitch

eitch:
Thanks for your explanation. That definitely sounds like a good reason. Then i hope you manage to get it working the way it is easier to debug.

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -312,29 +331,63 @@ public int write(byte b) {
...
+            txEntry.rx_buf = 0;
+            txEntry.bits_per_word = BITS8;
+            txEntry.speed_hz = config.baud();
+            txEntry.delay_usecs = 0;
+            txEntry.len = chunk.length ;
+            txEntry.cs_change = 1 ;
+            int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(1), txEntry);
+            logger.trace("[SPI::WRITE] <- Number bytes {} ", ret);
+            if (ret < 0) {
+                logger.error("Could not write SPI message. ret {}, error: {}", ret, Native.getLastError());
```
stefanhaustein, line 385: Shouldn't this be an exception?

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -312,29 +331,63 @@ public int write(byte b) {
...
-            length = 0;
+        Objects.checkFromIndexSize(offset, length, data.length);
+
+        int start = offset;
+        while (start < data.length) {
+            PeerAccessibleMemory buf = new PeerAccessibleMemory(SPI_BUFFSIZ);
+            spi_ioc_transfer txEntry = new spi_ioc_transfer() ;
+            int end = Math.min(data.length, start + SPI_BUFFSIZ);
+            byte[] chunk = Arrays.copyOfRange(data, start, end);
+            buf.write(0, chunk, 0 , chunk.length);
```
stefanhaustein, line 373: Why the chunk allocation here -- opposed to 
```
buf.write(0, data, start, end - start);
```

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -312,29 +331,63 @@ public int write(byte b) {
...
-
-        int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(1), transfer);
-        if (ret < 0) {
-            LOG.error("Could not write SPI message. ret {}, error: {}", ret, Native.getLastError());
-            length = 0;
+        Objects.checkFromIndexSize(offset, length, data.length);
+
+        int start = offset;
+        while (start < data.length) {
+            PeerAccessibleMemory buf = new PeerAccessibleMemory(SPI_BUFFSIZ);
```
stefanhaustein, line 369: Could this be pulled out of the loop? It looks like the contents are written in line 382

Old line 317 suggests to re-use it even more.....

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -312,29 +331,63 @@ public int write(byte b) {
...
+            spi_ioc_transfer txEntry = new spi_ioc_transfer() ;
+            int end = Math.min(data.length, start + SPI_BUFFSIZ);
+            byte[] chunk = Arrays.copyOfRange(data, start, end);
+            buf.write(0, chunk, 0 , chunk.length);
+            // set fields in transfer msg
+            txEntry.tx_buf = buf.getPeer();
```
stefanhaustein, line 375: Is it necessary to have all these repeated in the loop?

stefanhaustein:
> This implementation is same as found in pigpio It also creates multiple spi writes (ioctl call) I communicated with the pigpio author, them telling me the multi writes to get all the data into the spi device operated correctly.
> 
> I have another version of this fix written, it uses multi transfers in an array that is passed on a single ioctl. This would be a better solution. I hit intermittent IOERRORs when using this code. I will write a simple testcase to duplicate the error and post details to raspberryPi forum. If I use this code with a single entry in the array it works

Can you share this code?

> 
> I don't know how many user write more than 4k bytes. And they can increase the buffer if needed. Just this user needed 115200 bytes.

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpiProviderImpl.java
```diff
@@ -38,19 +38,21 @@
...
     implements LinuxFsSpiProvider {
 
+
     public LinuxFsSpiProviderImpl() {
         this.id = ID;
         this.name = NAME;
     }
 
-    @Override
+        @Override
```
stefanhaustein, line 55: This looks odd? Was this file included intentionally?

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -312,29 +331,63 @@ public int write(byte b) {
...
+            txEntry.rx_buf = 0;
+            txEntry.bits_per_word = BITS8;
+            txEntry.speed_hz = config.baud();
+            txEntry.delay_usecs = 0;
+            txEntry.len = chunk.length ;
+            txEntry.cs_change = 1 ;
```
stefanhaustein, line 381: I think this does the opposite of what we want here: It toggles the cs line to separate the transactions, but we want a single "logical" transaction.

For separate logical transactions, users are likely to use separate write calls already.

Imagine you send image data, and 4096 is by chance after the red value but before the green value of a single pixel... I don't think this would be the best time to toggle the cs line.

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -312,29 +331,63 @@ public int write(byte b) {
...
-        transfer.len = length;
-
-        int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(1), transfer);
-        if (ret < 0) {
-            LOG.error("Could not write SPI message. ret {}, error: {}", ret, Native.getLastError());
-            length = 0;
+        Objects.checkFromIndexSize(offset, length, data.length);
+
+        int start = offset;
+        while (start < data.length) {
```
stefanhaustein, line 368: Shouldn't this check be for (offset + length) instead of data.length?

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -312,29 +331,63 @@ public int write(byte b) {
...
+            PeerAccessibleMemory buf = new PeerAccessibleMemory(SPI_BUFFSIZ);
+            spi_ioc_transfer txEntry = new spi_ioc_transfer() ;
+            int end = Math.min(data.length, start + SPI_BUFFSIZ);
```
stefanhaustein, line 371: shouldn't this be 

```
int end = Math.min(offset + lenght, start + SPI_BUFFSIZ);
```

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: Most fails in the providers is a < 0 return code.    We could make this a V4 conversation whether to change to an exception in some failures. This is another case of history before most of us were involved.

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: done

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: multi spi_ioc_transfer version will need this, but agree wasn't needed in this version

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: multi spi_ioc_transfer version will need this, but agree wasn't needed in this version

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpiProviderImpl.java
taartspi: I think from a rebase operation

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: multi spi_ioc_transfer version will handle this where only last spi_ioc_transfer has flag set to release CS

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: you are correct the offset was not properly applied

PR #484, opened 2025-09-05
[Fix] Fix address uniqueness to use IO Type and address

eitch:
This should fix the issue where a PWM and GPIO use the same address as integer, but should be possible as they are different I/Os.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/registry/impl/DefaultRuntimeRegistry.java
+++ b/pi4j-core/src/main/java/com/pi4j/registry/impl/DefaultRuntimeRegistry.java
@@ -145,6 +145,7 @@ public class DefaultRuntimeRegistry implements RuntimeRegistry {
             throw new IONotFoundException(_id);
 
         IO shutdownInstance = instances.get(_id);
+        remove(shutdownInstance);
 
         // return the shutdown I/O provider instances
         return (T) shutdownInstance;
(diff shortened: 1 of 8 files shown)


pi4j-core/src/main/java/com/pi4j/registry/impl/DefaultRuntimeRegistry.java
```diff
@@ -187,8 +197,9 @@ public synchronized boolean exists(String id) {
     }
 
     @Override
-    public synchronized boolean exists(int address) {
-        return usedAddresses.contains(address);
+    public synchronized boolean exists(IOType ioType, int address) {
```
FDelporte, line 207: should this be unit tested?

FDelporte (approved):

eitch:
@FDelporte i added tests

FDelporte:
> added tests
@eitch thanks, looks good!

FDelporte:
> looks good!

or not? ;-)

```
Error:    RegistryTest.testCreateMultipleSameAddress:102 » IOAlreadyExists IO instance [DIN-3] already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.
Error:    RegistryTest.testShutdownAndRecreate:89 » IOAlreadyExists IO instance [DIN-3] already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.
```

eitch:
There was an issue with tests not cleaned up between each run. Is fixed now.

eitch:
@taartspi @FDelporte can we merge?

FDelporte:
> @taartspi @FDelporte can we merge?

LGTM

taartspi (approved):
Looks good

Issue #485, opened 2025-09-07
Shutting down GpioDDigitalOutput does not release the GPIO line


IAmNickNack:
I have a requirement to be able to remove a device at runtime and possibly to then re-add it at a later date.

I expected this to be possible thanks to the shutdown logic invoked when I remove the device from the registry.  However, this doesn't appear to release the GPIO line and only to be done in response to a shutdown of the context.

I can successfully add, use and then remove (from the registry) a DigitalOutput, but re-creating will fail with:

```
com.pi4j.exception.InitializeException: Failed to initialize output reset-pin
	at com.pi4j.plugin.gpiod.provider.gpio.digital.GpioDDigitalOutput.initialize(GpioDDigitalOutput.java:76) ~[pi4j-plugin-gpiod-3.0.2.jar:na]
	at com.pi4j.plugin.gpiod.provider.gpio.digital.GpioDDigitalOutput.initialize(GpioDDigitalOutput.java:46) ~[pi4j-plugin-gpiod-3.0.2.jar:na]
	at com.pi4j.registry.impl.DefaultRuntimeRegistry.add(DefaultRuntimeRegistry.java:95) ~[pi4j-core-3.0.2.jar:na]
	... 54 common frames omitted
Caused by: com.pi4j.library.gpiod.internal.GpioDException: c_gpiod_line_request_output failed: -1 (Device or resource busy)
	at com.pi4j.library.gpiod.internal.GpioD.lineRequestOutput(GpioD.java:156) ~[pi4j-library-gpiod-3.0.2.jar:na]
	at com.pi4j.library.gpiod.internal.GpioLine.requestOutput(GpioLine.java:70) ~[pi4j-library-gpiod-3.0.2.jar:na]
	at com.pi4j.plugin.gpiod.provider.gpio.digital.GpioDDigitalOutput.initialize(GpioDDigitalOutput.java:74) ~[pi4j-plugin-gpiod-3.0.2.jar:na]
	... 56 common frames omitted
```

One dirty workaround I have found for this is to explicitly release the line after I have removed the device from the registry: 

```java
var device = pi4j.registry().remove(id);
if (device instanceof GpioDDigitalOutput gpiod) {
    Field field = GpioDDigitalOutput.class.getDeclaredField("line");
    field.setAccessible(true);
    var line = (GpioLine)field.get(gpiod);
    GpioDContext.getInstance().closeLine(line);
}
```

... which is obviously not a viable, long term solution to my problem.

[...]

PR #486, opened 2025-09-07
Make GpioLine closeable

IAmNickNack:
This PR would address a possible issue described in #485 

`GpioLine` is made `Closeable` and on close the GpioD library to release the line.

GpioD input and output are updated to close the line during `shutdown`


IAmNickNack:
I will re-raise this PR as it is attributing the commit to the wrong email address

PR #487, opened 2025-09-07
Release GpioLine on IO device shutdown

IAmNickNack:
This PR could address a possible issue described in https://github.com/Pi4J/pi4j/issues/485

GpioLine is made Closeable and on close the GpioD library to release the line.

GpioD input and output are updated to close the line during shutdown


eitch:
Hi  @IAmNickNack thanks for the PR.

The close method interferes with com.pi4j.library.gpiod.internal.GpioDContext#close() - we store the GpioLine objects in a map and the object doesn't get removed from this map when we just call GpioLine.close(). Furthermore GpioDContext.close() doesn't call close on the line, but duplicates  the close implementation.

We need to update the implementation to remove the GpioLine from the map as well, so that it gets recreated properly.

eitch (changes requested):
See my comment

eitch (approved):
Much better, thanks!

PR #490, opened 2025-09-10
Gate digital output event creation on listner / binding presence

stefanhaustein:
Also print the plugin ids in debug output as this is what they are referred to by in other places (such as .plugin in builders)

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/event/EventManager.java
+++ b/pi4j-core/src/main/java/com/pi4j/event/EventManager.java
@@ -43,6 +43,11 @@ public class EventManager<SOURCE_TYPE, LISTENER_TYPE extends Listener, EVENT_TYP
         this.delegate = delegate;
     }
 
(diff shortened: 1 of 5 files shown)


pi4j-core/src/main/java/com/pi4j/event/EventManager.java
```diff
@@ -43,6 +43,11 @@ public EventManager(SOURCE_TYPE source, EventDelegate<LISTENER_TYPE,EVENT_TYPE>
         this.delegate = delegate;
     }
 
+    /** Returns true if this event manage has at least one listener. Used to skip event creation otherwise. */
```
FDelporte: typo "manager"

FDelporte (approved):

pi4j-core/src/main/java/com/pi4j/event/EventManager.java
stefanhaustein: Done

eitch (approved):
looks good

PR #491, opened 2025-09-10
PWM IO Change Float to Integer and remove float related conversions

taartspi:
Used PWN testcase on Pi5 to use the PWM function with a LED

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/pwm/Pwm.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/pwm/Pwm.java
@@ -150,19 +150,19 @@ public interface Pwm extends IO<Pwm, PwmConfig, PwmProvider>, OnOff<Pwm> {
      * Turn the PWM signal [ON] using a specified duty-cycle (%)
      * at the pre-configured frequency (Hz).
      *
-     * @param dutyCycle  The duty-cycle value is a decimal value that represents the
+     * @param dutyCycle  The duty-cycle value is an Integer value that represents the
      *                   percentage of the ON vs OFF time of the PWM signal for each
      *                   period.  A value of 50 represents a duty-cycle where half of
      *                   the time period the signal is LOW and the other half is HIGH.
-     *                   The duty-cycle range is valid from 0 to 100 including factional
-     *                   values. (Values above 50% mean the signal will remain HIGH more
+     *                   The duty-cycle range is valid from 0 to 100.
+     *                   (Values above 50% mean the signal will remain HIGH more
      *                   time than LOW.)
      * @return returns this PWM instance
      * @throws IOException if fails to communicate with the PWM pin
      */
-    default Pwm on(Number dutyCycle) throws IOException{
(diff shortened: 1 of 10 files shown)


pi4j-core/src/main/java/com/pi4j/io/pwm/Pwm.java
```diff
@@ -204,19 +204,19 @@ default Pwm on(Number dutyCycle, int frequency) throws IOException{
     }
 
     /**
-     *  Get the duty-cycle value as a decimal value that represents the
-     *  percentage of the ON vs OFF time of the PWM signal for each
-     *  period.  The duty-cycle range is valid from 0 to 100 including
-     *  factional values. (Values above 50% mean the signal will remain
-     *  HIGH more time than LOW.)
+     * Get the duty-cycle value as a decimal value that represents the
```
FDelporte: Description "decimal value" and return type Integer don't match... I'm OK with the Integer if that correctly represents the value, but should match the JavaDoc...

pi4j-core/src/main/java/com/pi4j/io/pwm/Pwm.java
```diff
@@ -204,19 +204,19 @@ default Pwm on(Number dutyCycle, int frequency) throws IOException{
...
+     * the time period the signal is LOW and the other half is HIGH.
      *
      * @return duty-cycle value expressed as a percentage (rage: 0-100)
      * @throws IOException if fails to communicate with the PWM pin
      */
-    float getDutyCycle() throws IOException;
+    Integer getDutyCycle() throws IOException;
 
     /**
      *  Get the duty-cycle value as a decimal value that represents the
```
FDelporte: Idem, also in other places...

pi4j-core/src/main/java/com/pi4j/io/pwm/impl/DefaultPwmConfig.java
```diff
@@ -93,7 +93,7 @@ protected DefaultPwmConfig(Map<String,String> properties){
 
         // load optional pwm duty-cycle from properties
         if(properties.containsKey(DUTY_CYCLE_KEY)){
-            this.dutyCycle = Float.parseFloat(properties.get(DUTY_CYCLE_KEY));
+            this.dutyCycle = Integer.parseInt(properties.get(DUTY_CYCLE_KEY));
```
FDelporte, line 96: Would be better to have a generic method, e.g. properties.getInt(string, int default), that has exception handling if the value can't be converted to integer and return the default value in such case

FDelporte (changes requested):
OK with the changes, but should align with JavaDoc...

pi4j-core/src/main/java/com/pi4j/io/pwm/impl/DefaultPwmConfig.java
FDelporte: I looked through the code, and this same approach is used in a lot of places. Something to rework later...

PR #492, opened 2025-09-22
Move PWM chip detection to BoardInfo

FDelporte:
Related to Slack discussion:

```
Is it something we need to solve in FFM API? It sounds like we need to provide better API and suggestions with e.g. BoardInfoHelper but not in provider's code?
```

What do you think of this proposal to move PWM Chip detection to BoardInfo?
So if needed, can also be used for FFM API initialization.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/boardinfo/util/PwmChipUtil.java
+++ b/pi4j-core/src/main/java/com/pi4j/boardinfo/util/PwmChipUtil.java
@@ -1,5 +1,4 @@
-package com.pi4j.plugin.linuxfs.provider.pwm;
-
+package com.pi4j.boardinfo.util;
 
 /*
  *
@@ -30,9 +29,7 @@ package com.pi4j.plugin.linuxfs.provider.pwm;
  *
  */
 
-
 import com.pi4j.boardinfo.util.command.CommandResult;
-import com.pi4j.plugin.linuxfs.internal.LinuxPwm;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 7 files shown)


pi4j-core/src/main/java/com/pi4j/boardinfo/util/PwmChipUtil.java
```diff
@@ -59,11 +76,9 @@ public class LinuxFsPwmUtil {
      * @return Int value for the pwmchip if configured,
      * else default to original expected value of 2.
      */
-
-
     public static int getPWMChipForRP1(String pwmFileSystemPath) {
-        Logger logger = LoggerFactory.getLogger(LinuxFsPwmUtil.class);
-        int pwmChip = LinuxPwm.DEFAULT_RP1_PWM_CHIP;
+        Logger logger = LoggerFactory.getLogger(PwmChipUtil.class);
+        int pwmChip = DEFAULT_RP1_PWM_CHIP;
```
eitch, line 82: can we please change the ls -l to use the Files.walk API: https://gist.github.com/eitch/5079aa833991715e774c1e4e6777b89d

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/internal/LinuxPwm.java
```diff
@@ -194,10 +182,13 @@ public Polarity polarity() throws IOException {
      */
     public Polarity getPolarity() throws IOException {
         var path = Paths.get(pwmPath, "polarity");
-        switch(Files.readString(path).trim().toLowerCase()){
-            case "inversed": return Polarity.INVERSED;
-            case "normal": return Polarity.NORMAL;
-            default: return Polarity.UNKNOWN;
+        switch (Files.readString(path).trim().toLowerCase()) {
```
eitch: how about we switch to the new switch expressions? Or is that not in the current JDK?

eitch (commented):
I have a few comments. Otherwise looks good

taartspi (approved):
Thanks Frank

pi4j-core/src/main/java/com/pi4j/boardinfo/util/PwmChipUtil.java
FDelporte: done, @taartspi can you verify the improved code please?

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/internal/LinuxPwm.java
FDelporte: good remark, done

pi4j-core/src/test/java/com/pi4j/boardinfo/util/PWMChipUtilTest.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.List;
+import java.util.Optional;
+
+import static com.pi4j.boardinfo.util.PwmChipUtil.parsePWMPaths;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class PWMChipUtilTest {
+
+    @Test
+    void shouldReturnCorrectPWMChipForRP1Paths() {
```
eitch, line 14: When using Files.walk() then this test won't work anymore.

taartspi (approved):

pi4j-core/src/test/java/com/pi4j/boardinfo/util/PWMChipUtilTest.java
FDelporte: yes it does, as the functionality is done in two steps exactly to be able to test. first get a list of paths, and secondly get the PWM address from that list.

FDelporte: I refactored to these two steps on purpose, as it was unclear how the address was retrieved from the paths. Should be more clear (and testable) with this approach...

taartspi:
The latest updates look good

PR #493, opened 2025-09-23
the change from using ls -l provided less info to the utility. Previo…

taartspi:
…ulsy the pwmchip number was found early in the path detail and at the end. Now it is only at the end of the path.  The code as it was walked off the end of the array.  The fix a simple break.


taartspi:
Updated the testcase so the data looks the same as in the runtime.   
Used a sledge hammer to drive tacks to fix this.  Too late at night to rewrite the method for a cleaner fix and won't have time until Wednesday the earliest    Tom   
 @eitch @FDelporte

FDelporte (approved):

PR #499, opened 2025-09-26
Add a listenable OnOffRead

stefanhaustein:
When combining digital driver inputs with IO expanders, we need a way to hand over a representation of the expander pins to the driver.

Also move setState(boolean) up to OnOffWrite for convenience (ideally this would be the one non-default method, so only one override is needed in implementations, but I guess this ship has sailed now....)

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/ListenableOnOffRead.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/ListenableOnOffRead.java
@@ -40,7 +40,10 @@ import java.util.function.Consumer;
  */
 public interface ListenableOnOffRead<T> extends OnOffRead<T> {
 
-    /** Adds a consumer as a listener to this object. */
+    /**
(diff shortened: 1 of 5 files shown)


FDelporte (approved):

stefanhaustein:
@FDelporte , do you mind if I change this as outlined in the slack drivers channel to avoid wrapping DigitalInput?

eitch:
Looks good. Go ahead and change it. I think we can be experimental a bit now with 4.0, no?

FDelporte:
Agree with both of you 😀

stefanhaustein:
Ok, I definitely do like this version better, as it allows me to "just" replace DigitalInput with ListenableOnOffRead without wrapping. The adder name is a bit unfortunate, but I think just using the argument type name is ok-ish and doesn't throw in yet another name.

stefanhaustein:
If you are ok with this, can somebody merge? I have no powers here O:)

Issue #500, opened 2025-09-27
Pi4J.newAutoContext() no longer works in 4.0.0-SNAPSHOT build for gpiod


IAmNickNack:
On Pi5, `newAutoContext` worked in previous builds. E.g. `3.1.0-SNAPSHOT` didn't have this problem.

`GpioDContext` appears not to mind `chipName` being null. `DefaultContextBuilder` now defaults `gpio.chip.name` to `unknown` and causes any chip not to be found.

https://github.com/Pi4J/pi4j/commit/55bead2b4f32e981d0aae51d1bdf72e8b671302e#diff-9d62545857e9677be136acfdc496571edced749bea4f39aeef2a7e6695723c5fR307

```
java.lang.IllegalStateException: Couldn't find gpiochip with name: unknown
	at com.pi4j.library.gpiod.internal.GpioDContext.initialize(GpioDContext.java:132) ~[pi4j-library-gpiod-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.plugin.gpiod.provider.gpio.digital.GpioDDigitalInputProviderImpl.initialize(GpioDDigitalInputProviderImpl.java:48) ~[pi4j-plugin-gpiod-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.plugin.gpiod.provider.gpio.digital.GpioDDigitalInputProviderImpl.initialize(GpioDDigitalInputProviderImpl.java:14) ~[pi4j-plugin-gpiod-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.provider.impl.DefaultRuntimeProviders.initializeProvider(DefaultRuntimeProviders.java:273) ~[pi4j-core-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.provider.impl.DefaultRuntimeProviders.add(DefaultRuntimeProviders.java:249) ~[pi4j-core-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.provider.impl.DefaultRuntimeProviders.add(DefaultRuntimeProviders.java:229) ~[pi4j-core-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.provider.impl.DefaultRuntimeProviders.initialize(DefaultRuntimeProviders.java:354) ~[pi4j-core-4.0.0-SNAPSHOT.jar:na]
[...]

PR #501, opened 2025-09-27
Move default chip check to FFM providers

IAmNickNack:
This PR suggests an alternative approach for the FFM plugin to read a default value for `gpio.chip.name` so as not to break existing functionality as described in #500 

If necessary, maybe changing the "default values" behaviour in shared code can be addressed by a separate issue and PR instead

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/context/impl/DefaultContextBuilder.java
+++ b/pi4j-core/src/main/java/com/pi4j/context/impl/DefaultContextBuilder.java
@@ -304,7 +304,6 @@ public class DefaultContextBuilder implements ContextBuilder {
 
             @Override
             public Map<String, String> properties() {
-                builder.properties.put("gpio.chip.name", gpioChipName != null ? gpioChipName : "unknown");
                 return Collections.unmodifiableMap(builder.properties);
             }
         };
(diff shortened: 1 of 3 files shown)


FDelporte:
Can this be integrated in boardinfo somehow to make it more generic? Probably different for rpi 5 versus earlier versions?

IAmNickNack:
The board info approach would probably be the ideal, yes. 

This PR is primarily to highlight the issue and fix the current implementation if necessary. It simply reverts a change to the common code and localises the requirement to have a default value to where a default value is wanted. 

I've not had to touch the board info APIs yet, so not familiar with it. I'd be happy to take a look, but wouldn't be able to commit to a timeline for a PR.

DigitalSmile:
As for the fix looks good to me. But I would consider bring every bit of logic about chip numbers or hardware somewhere else (e.g. boardinfo class or similar).

DigitalSmile (approved):

stefanhaustein:
Would it make sense to change the settings to auto-commit? 

It looks like there are at least two PRs pending here that are good-to-go but not merged, causing issues down the line....

FDelporte (approved):

PR #503, opened 2025-09-27
Remove the generic type for ListenableOnOff.Impl

stefanhaustein:
As this is a concrete class that doesn't make sense to subclass, we can use its type for the generic type parameter of OnOffRead.


FDelporte (approved):

stefanhaustein:
Can you also merge? I have no powers here O:)

FDelporte (approved):

Issue #504, opened 2025-09-27
FFM Digital Output fails to initialise when initial state is provided


IAmNickNack:
`DigitalOutputFFM` attempts to set the initial state before the device is initialised. This causes the `ioctl` call to fail because `chipFileDescriptor` is zero.

```
com.pi4j.exception.Pi4JException: com.pi4j.exception.Pi4JException: Error during call to method 'call' with data '[0, 3222320143, LineValues{bits=1, mask=1}]': Inappropriate ioctl for device (25)
	at com.pi4j.plugin.ffm.providers.gpio.DigitalOutputFFM.state(DigitalOutputFFM.java:102) ~[pi4j-plugin-ffm-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.io.gpio.digital.DigitalOutputBase.initialize(DigitalOutputBase.java:66) ~[pi4j-core-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.plugin.ffm.providers.gpio.DigitalOutputFFM.initialize(DigitalOutputFFM.java:43) ~[pi4j-plugin-ffm-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.plugin.ffm.providers.gpio.DigitalOutputFFM.initialize(DigitalOutputFFM.java:24) ~[pi4j-plugin-ffm-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.registry.impl.DefaultRuntimeRegistry.add(DefaultRuntimeRegistry.java:97) ~[pi4j-core-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.registry.impl.DefaultRegistry.add(DefaultRegistry.java:98) ~[pi4j-core-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.plugin.ffm.providers.gpio.DigitalOutputFFMProviderImpl.create(DigitalOutputFFMProviderImpl.java:25) ~[pi4j-plugin-ffm-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.plugin.ffm.providers.gpio.DigitalOutputFFMProviderImpl.create(DigitalOutputFFMProviderImpl.java:12) ~[pi4j-plugin-ffm-4.0.0-SNAPSHOT.jar:na]
	at com.pi4j.context.Context.create(Context.java:363) ~[pi4j-core-4.0.0-SNAPSHOT.jar:na]
[...]

PR #505, opened 2025-09-27
Call `super.initialize` after initialisation

IAmNickNack:
This PR provides a fix for #504 to allow initial state of the output to be set during initialisation


FDelporte:
This is only possible since Java 25, correct?
Can this be unit tested?

IAmNickNack:
If this were in the constructor, yes it would only be possible with java 25.  This is in the `initialize` method, so compatible with older versions. 

I have added a unit test in the style of the other GPIO integration tests. I see this test class is currently `@Disabled`, but have run it locally. It does fail on shutdown, but I believe that's because I don't have my pi configured to the required extent: 

```
Probably you need to add the GPIO Simulator bash script to sudoers file with visudo:
```

PR #513, opened 2025-09-29
Make IO closeable and add a base implementation to BaseIO

stefanhaustein:
Not exactly sure what the best place would be to avoid cycles with stuff that already implements close (maybe that's not a problem because then that will override anyway and we could push this up all the way to IOBase)?

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/IOBase.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/IOBase.java
@@ -91,12 +91,16 @@ public abstract class IOBase<IO_TYPE extends IO, CONFIG_TYPE extends IOConfig, P
      * Closes the driver by calling this.context().shutdown(this.getId()), which in turn calls
      * the local shutdown() method here via DefaultRuntimeRegistry.remove().
      * <p>
-     * IO implementations need to override the local shutdown method with implementation-
-     * specific shutdown behaviour as needed.
+     * Basically, for Pi4J, this constitutes an idempotent user convenience method for
+     * this.context().shutdown(this.getId())
+     * <p>
+     * Subclasses should typically override the local shutdown method with implementation-specific shutdown
+     * behaviour. Behaviour added here will not be triggered by context.shutdown()
      */
     @Override
-    public final void close() {
-        // Account for contextless tests or somehow just closing without initializing
+    public void close() {
+        // The null check accounts for contextless tests or somehow just closing without initializing
+        // The closed check ensures idempotency, as required by the close method contract.
         if (this.context != null && !closed) {
             this.closed = true;
             this.context.shutdown(getId());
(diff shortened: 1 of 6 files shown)


eitch:
This looks good. If we can move it up to IOBase then sure.

stefanhaustein:
On which interfaces would we want to expose this?

Wrt tests, I think we could just replace the corresponding shutdown calls in tests?

eitch:
Yes, i think we can just update the existing tests. No need to write new ones, if the existing ones need to be updated anyhow.

eitch:
Is IOBase a good place? From the looks of it, i would say yes. Doesn't make sense higher up. Lower down neither.

pi4j-core/src/main/java/com/pi4j/io/IOBase.java
```diff
@@ -104,6 +116,8 @@ public IO_TYPE initialize(Context context) throws InitializeException {
     /** {@inheritDoc} */
     @Override
     public IO_TYPE shutdown(Context context) throws ShutdownException {
+        // Avoid close calling shutdown after shutdown was called by other means
+        this.closed = true;
```
eitch, line 132: What if just thus method was called? Existing code would be doing that...

stefanhaustein: That's fine. Existing code is aware that shutdown can't be called multiple times. This just makes sure that calls to close won't try a shutdown again if shutdown was already called by other means. Perhaps one could argue that close strictly only needs to be idempotent with itself.... This will just prevent problems with people interpreting idempotency as "it's always safe to call close" (which might be the spirit behind requiring idempotency) if there is some inner pre-existing cleanup already.

eitch: True, but if the user just called this, then the IO wouldn't be shutdown, right? I.e. shouldn't we be calling close() here?

stefanhaustein: I was assuming that this is an unfortunately named internal method called by context.shutdown(io)...?

eitch: Ok, i checked the code again, and what happens is that the registry calls this method:
com.pi4j.registry.impl.DefaultRuntimeRegistry#remove(T)

So if we add IOBase.close(), when is this called from the Pi4j context? Ideally close() and shutdown() should do the same, no?

stefanhaustein: Close and shutdown can't be the same without significantly changing the architecture. 

Currently, the call sequence is 

> context.shutown(io.id) -> registry.remove(io) -> io.shutdown(context)

Close just introduces a convenience call for context.shutdown(this), i.e. the first call in this sequence.

The context/registry need to be able to shut down the instance without leading to endless recursion. 

This could have been done in a different way by letting close trigger the de-registration without doing the removal. Basically then the registry would just call io.close() and io would call something like context.remove(this). But that's not how it is and I don't think we should change this right now.

eitch: Ok, makes sense

eitch (approved):

PR #515, opened 2025-10-01
#508 Fixes for Permission Checker

### LINKED ISSUE #508: FFM PermissionHelper doesn't work on PI5 Bookworm
File permission checks fail on my device. 

PermissionHelper assumes the `gpio` group on Debian which fails on what is (to my knowledge) the default setup on Pi5

```
➜  ~ uname -a
Linux rpi-5 6.12.34+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.12.34-1+rpt1~bookworm (2025-06-26) aarch64 GNU/Linux
➜  ~ ls -l /dev/gpiochip0
crw-rw---- 1 root dialout 254, 0 Sep 27 13:17 /dev/gpiochip0
➜  ~ ls -l /dev/i2c-1
crw-rw---- 1 root i2c 89, 1 Sep 27 13:17 /dev/i2c-1
➜  ~ ls -l /dev/spidev0.0
crw-rw---- 1 root spi 153, 0 Sep 27 13:17 /dev/spidev0.0
```

Additionally, the helper is applied to i2c and SPI, but doesn't check the device type and assumes these should also be in the `gpio` group.

https://github.com/Pi4J/pi4j/blob/develop/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/common/PermissionHelper.java#L103-L112

FFM appears to work once the check is removed. 

Should the check be fixed or removed?

DigitalSmile:


plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/common/PermissionHelper.java
```diff
@@ -80,18 +108,28 @@ public static void checkUser() {
      * Checks device permissions to run the provider.
      * - check device existence. Throws exception
      * - checks group `dialout`/`gpio` and group permissions to write/read. Throws exception.
+     *
      * @param devicePath path to be checked
      */
     public static void checkDevice(String devicePath) {
         var path = Paths.get(devicePath);
+        // checking that device is physically exists
         if (!path.toFile().exists()) {
```
eitch, line 134: so after this check i would simply check if we have write access, then if we don't we can give additional checks a go. A user might give access with a completely different group through udev, in which case all this will fail again.

DigitalSmile:
I have implemented PermissionChecker to lazily on provider / device creation make sure everything work as expected.
Checks:
- run with sudo / root
- no relative group present in the system (gpio/dialout, i2c, spi, input)
- user is not member of this groups
- device does not exists
- device has no group permissions listed above
- device has excessive permissions in others

I have wrote simple explanation in error/exception description to help user with the issue. When the documentation page will be available I will put relevant links and adjust writings.

Users with current implementation should not be affected, if they already work with devices (meaning all permissions are set up).

eitch:
There are many checks, and it seems sensible. With out testing it'll be difficult to say how well it works. But i do have a few questions:
- I2C: Why not just check if write access is available to the file and then stop, otherwise do the check and write the output depending on what is expected?
- The native downcalls to setgrent, etc. Since we're thinking of pi4j v4.x to be not necessarily just for the Raspberry Pi, can we expect glibc is to always be there? Is this POSIX? Just asking so we don't shoot ourselves again in the foot for something that might work but the check fails.

DigitalSmile:
> There are many checks, and it seems sensible. With out testing it'll be difficult to say how well it works.

If a user is already working with hardware, than it supposed to work the same (if you have no access to device, how it is working now?). Except a couple of warnings in logs, it is not blocking user from using pi4j.

> I2C: Why not just check if write access is available to the file and then stop, otherwise do the check and write the output depending on what is expected?

Good point, but how to achieve this? We can use `access` glibc function to check, but logically it will be the same as written in my code. You cannot know if user has real access before start writing/reading. That is why my idea is to make all possible checks beforehand to ensure it is safe and does not have "Permission denied" errors.

> The native downcalls to setgrent, etc. Since we're thinking of pi4j v4.x to be not necessarily just for the Raspberry Pi, can we expect glibc is to always be there? Is this POSIX? Just asking so we don't shoot ourselves again in the foot for something that might work but the check fails.

JVM itself is cannot run without glibc :) So we are in a safe harbour. Specifically, `setgrent` and others in this PR are POSIX functions (https://man7.org/linux/man-pages/man3/getgrent.3.html). They are quite popular and are implemented in most of distros.
As for the other functions in ffm-api, I am double checking with usage, so unless we are not supporting some exotic OS (like QNX) we are fine.

eitch:
> Good point, but how to achieve this? We can use `access` glibc function to check, but logically it will be the same as written in my code. You cannot know if user has real access before start writing/reading. That is why my idea is to make all possible checks beforehand to ensure it is safe and does not have "Permission denied" errors.

Till now i was certain that if i did a simple if(new File().canWrite()) {} then i am safe. Do you have reason to believe this is not the case?

For the rest i am ok with.

DigitalSmile:
> Till now i was certain that if i did a simple if(new File().canWrite()) {} then i am safe. Do you have reason to believe this is not the case?

From javadoc:
```
Tests whether the application can modify the file denoted by this abstract pathname. On some platforms it may be possible to start the Java virtual machine with special privileges that allow it to modify files that are marked read-only. Consequently, this method may return true even though the file is marked read-only.
```

Essentially, the function return permissions accessed only by application. But we are using things, such as `ioctl` to access hardware with kernel syscalls. And what is written in FS is not always corresponding to user having access rights to hardware through file. For example, that's usually the case for `configfs` - you can catch "Permission denied" while writing/reading, even if you are running with `root` or `sudo`

eitch:
Ok, i guess then we should merge this PR =)

eitch (approved):

DigitalSmile:
We can reiterate further on with enhancement features. E.g. with custom user groups, which are not following distro guidelines.

PR #517, opened 2025-10-03
Rename the internal shutdown method in Lifecycle to shutdownInternal

stefanhaustein:
The intention is to avoid user confusion: there were several shutdown-related bugs where users were calling io.shutdown(ctx) directly -- instead of correctly calling ctx.shutdown(io) 

CI doesn't seem to like it currently but let's have a discussion if we want this before I sink more time into this O:)

### DIFF
--- a/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/DigitalInputFFM.java
+++ b/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/DigitalInputFFM.java
@@ -13,6 +13,7 @@ import com.pi4j.plugin.ffm.common.gpio.PinEvent;
 import com.pi4j.plugin.ffm.common.gpio.PinEventProcessing;
 import com.pi4j.plugin.ffm.common.gpio.PinFlag;
 import com.pi4j.plugin.ffm.common.gpio.enums.LineAttributeId;
+import com.pi4j.plugin.ffm.common.gpio.structs.*;
 import com.pi4j.plugin.ffm.common.ioctl.Command;
 import com.pi4j.plugin.ffm.common.ioctl.IoctlNative;
 import com.pi4j.plugin.ffm.common.poll.PollFlag;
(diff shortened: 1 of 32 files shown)


eitch (approved):
Looks good to me as is as discussed

eitch:
But the build fails, so something is not right.

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/DigitalInputFFM.java
```diff
@@ -13,7 +13,6 @@
 import com.pi4j.plugin.ffm.common.gpio.PinEventProcessing;
 import com.pi4j.plugin.ffm.common.gpio.PinFlag;
 import com.pi4j.plugin.ffm.common.gpio.enums.LineAttributeId;
-import com.pi4j.plugin.ffm.common.gpio.structs.*;
```
DigitalSmile, line 16: Build fails because you removed imports

stefanhaustein: Thanks! 

I wonder why they got dropped -- I guess either intellij or checkstyle seems to think they are not needed.

IntelliJ also complained about the language level, and when I increased it, it added something to the pom file... Perhaps there is something else we need to set apart from "25" in the pom? I'll try to reproduce this separately....

eitch: IntelliJ always has issues when the Java version is lifted in the pom.xml. just revert the changes in the pom.xml

stefanhaustein: I had done that but now I am wondering if it would be safer to set the language level there, too (like intellij wanted to)

DigitalSmile (approved):

Issue #516, opened 2025-10-03
DeferredDelay


stefanhaustein:
Would you mind if I add a class implementing this API to com.pi4j.util:

```
/**
 * Device communication often requires delays between different operations.
 * Just sleeping often will waste this time. For instance, while waiting for 
 * a display to be ready to receive the next image, the next image could be
 * rendered already. This class allows setting a delay after an operation imposing
 * a delay -- but the delay is materialized only later -- typically, immediately before the
 * next IO operation. 
 */
class DeferredDelay {
   public void setDelayMillis(int millis);
   public void setDelayNanos(int nanos);
   public Instant getBusyUntil();

   /** Sleeps until the delay time requested by setDelay has passed. */ 
   public void materializeDelay();
}
```

FDelporte:
Seems a good idea

eitch:
+1 from me too.

stefanhaustein:
Addressed in https://github.com/Pi4J/pi4j/pull/518

PR #518, opened 2025-10-03
Add DeferredDelay

stefanhaustein:
Addresses  #516

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/util/DeferredDelay.java
+++ b/pi4j-core/src/main/java/com/pi4j/util/DeferredDelay.java
@@ -1,0 +1,71 @@
+package com.pi4j.util;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+
+/**
+ * Provide a way to set a sleep time -- and to materialize it later at a different place, where the sleep time
+ * is actually required.
+ * <p>
+ * Device communication often requires delays between different operations. Just sleeping often will waste this time.
+ * For instance, while waiting for a display to be ready to receive the next image, the next image could be
+ * rendered already.
+ * <p>
+ * This class allows setting a delay after an operation imposing a delay -- but then defer its materialized until the
+ * delay is actually required -- typically, immediately before the next IO operation.
+ */
+public class DeferredDelay {
+
+    private Instant busyUntil = Instant.now();
+
+    /**
+     * Sets a delay in milliseconds that will be materialized when materializeDelay() is called.
+     * <p>
+     * If a delay was set already previously, it will only be overridden if it results in a later "busyUntil" time.
+     * <p>
+     * Typically called after device IO calls where it's known how long command processing will take and the
+     * device will be ready again.
+     */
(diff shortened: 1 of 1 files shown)


taartspi (approved):
Thank you sir

DigitalSmile (commented):
I am sorry, I don't get the idea...
Can you please provide an example of usage? It seems, that time will be wasted anyway, because of thread locking and switching contexts (if we talk about nanos and not millis). Where you can save time?

stefanhaustein:
Devices typically have some time they are busy with processing a command before they can accept new commands, just take a look at the specs for bmp280, bme280, scd4x, hd44780.... Another similar case is WS2812b, which determines if data goes to the next led in the chain or restarts at the beginning using a delay of 500 microseconds IIRC.

So one might implement something like this to make sure this timing is honoured:

```
public void gameLoop() {
  while (true) {
    sendImageToDisplay();
    Thread.sleep(20);  // Display needs 20ms before it can accept new commands (= the next call to sendImageToDisplay())
    renderNewImage();  // Rendering takes 15ms on average
  }
}
```

As rendering time may vary, it's not safe to just subtract it from the delay. So the total delay here typically amounts to 35 ms, Resulting in a frame rate of ~28.

The idea here is to improve this as follows:

```
public void gameLoop() {
  while (true) {
    deferrredDelay.materializeDelay();  // This only delays by the remaining delay
    sendImageToDisplay();
    deferredDelay.setDelayMillis(20);  // Display needs 20ms before it can accept new commands
    renderNewImage();  // Rendering takes 15ms on average
  }
}
```

Resulting in a typical delay of 20ms (= ~50 fps)

stefanhaustein:
Can somebody merge? I have no powers here O:)

DigitalSmile:
> Devices typically have some time they are busy with processing a command before they can accept new commands, just take a look at the specs for bmp280, bme280, scd4x, hd44780.... Another similar case is WS2812b, which determines if data goes to the next led in the chain or restarts at the beginning using a delay of 500 microseconds IIRC.
> 
> So one might implement something like this to make sure this timing is honoured:
> 
> ```
> public void gameLoop() {
>   while (true) {
>     sendImageToDisplay();
>     Thread.sleep(20);  // Display needs 20ms before it can accept new commands (= the next call to sendImageToDisplay())
>     renderNewImage();  // Rendering takes 15ms on average
>   }
> }
> ```
> 
> As rendering time may vary, it's not safe to just subtract it from the delay. So the total delay here typically amounts to 35 ms, Resulting in a frame rate of ~28.
> 
> The idea here is to improve this as follows:
> 
> ```
> public void gameLoop() {
>   while (true) {
>     deferrredDelay.materializeDelay();  // This only delays by the remaining delay
>     sendImageToDisplay();
>     deferredDelay.setDelayMillis(20);  // Display needs 20ms before it can accept new commands
>     renderNewImage();  // Rendering takes 15ms on average
>   }
> }
> ```
> 
> Resulting in a typical delay of 20ms (= ~50 fps)

That makes sense now, thanks.

Issue #522, opened 2025-10-06
V4 I2C cleanup opportunity: Let's add default implementations for the register methods in I2c that don't already have one


stefanhaustein:
As far as I understand, registers are just common conventions -- they are not actually reflected in the protocol.

So we could just add default implementations, and providers wouldn't need to implement register support, reducing opportunities for unexpected differences or errors.

I am happy to create a PR to implement this if people agree that we should do this.

eitch:
Which methods do you think are missing? But yes, always a good idea to cover the use-cases

stefanhaustein:
I have tried to illustrate this here: https://github.com/Pi4J/pi4j/pull/524

stefanhaustein:
PS. Running out of forks, so here is the suggestion for the corresponding additions to `I2CBase.java`

```
   @Override
    public int readRegister(int register) {
        return execute(() -> {
            write(register);
            return read();
        });
    }Expand commentComment on line R92ResolvedCode has comments. Press enter to view.

    @Override
    public int readRegister(byte[] register, byte[] buffer, int offset, int length) {
        return writeRead(register, register.length, 0, buffer, length, offset);
    }

    @Override
    public int readRegister(int register, byte[] buffer, int offset, int length) {
        return execute(() ->{
            write(register);
            return read(buffer, offset, length);
        });
    }Expand commentComment on line R105ResolvedExpand commentComment on line R105ResolvedCode has comments. Press enter to view.

    @Override
    public int writeRegister(int register, byte b) {
        return write((byte) register, b);
    }

    @Override
    public int writeRegister(int register, byte[] data, int offset, int length) {
        Objects.checkFromIndexSize(offset, length, data.length);
        byte[] tmp = new byte[length + 1];
        tmp[0] = (byte) register;
        System.arraycopy(data, offset, tmp, 1, length);
        return write(tmp);
    }

    @Override
    public int writeRegister(byte[] register, byte[] data, int offset, int length) {
        Objects.checkFromIndexSize(offset, length, data.length);
        byte[] tmp = new byte[length + register.length];
        System.arraycopy(register,0, tmp, 0, register.length);
        System.arraycopy(data, offset, tmp, register.length, length);
        int rc = write(tmp);
        return (rc - register.length);  // do not include the register bytes as what was written...
    }
```

stefanhaustein:
[...]

PR #524, opened 2025-10-07
Implement non-default I2CRegisterDataReader / Writer method in I2CBase

stefanhaustein:
To my understanding, reading from a register / writing to a register is just writing the register address before the "actual" i2c read/write operation.

So we should be able to provide default implementations in IOBase, avoiding that all provider implementations need to implement the same utilities above the "raw" IO level. 

This was triggered by seeing the register based methods in the i2c gprc protobuf spec of @IAmNickNack's implemetation. 

I have taken the write operations from the linuxfs implementation, including the pre-existing return value inconsitency.

Reading the documentation of the I2CRagisterDataReader methods, I am not 100% sure there isn't something more to it -- people more familiar with I2C will probably know better. If so, I think we still should still try to model this as close to an "atomic" level as possible, reducing the methods to be implemented by providers (providing opportunities for differences in behaviour) as much as possible for v4.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBase.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBase.java
@@ -30,6 +30,7 @@ import com.pi4j.exception.ShutdownException;
 import com.pi4j.io.IOBase;
 import com.pi4j.io.i2c.impl.DefaultI2CRegister;
 
+import java.util.Objects;
 import java.util.concurrent.Callable;
 
 /**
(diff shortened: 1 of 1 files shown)


pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBase.java
```diff
@@ -82,6 +83,48 @@ public I2CRegister getRegister(int address) {
         return new DefaultI2CRegister(this, address);
     }
 
+    @Override
+    public int readRegister(int register) {
+        write(register);
+        return read();
+    }
```
taartspi, line 92: I believe this will be two operations to the i2c chip.   
Start w. Byte stop. Start r. Byte stop.   Implemented in ioctl in linuxfs. 
start w byte start r byte stop.  Single operation

pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBase.java
```diff
@@ -82,6 +83,48 @@ public I2CRegister getRegister(int address) {
...
+
+    @Override
+    public int readRegister(byte[] register, byte[] buffer, int offset, int length) {
+        write(register);
+        return read(buffer, offset, length);
```
taartspi: Same as above

taartspi: Same as above

taartspi: Same as above

pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBase.java
```diff
@@ -82,6 +83,48 @@ public I2CRegister getRegister(int address) {
...
+    }
+
+    @Override
+    public int readRegister(int register, byte[] buffer, int offset, int length) {
+        write(register);
+        return read(buffer, offset, length);
+    }
```
taartspi, line 105: Same as above

taartspi, line 105: Same as above

taartspi (changes requested):

pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBase.java
stefanhaustein: Better?

stefanhaustein:
I have switched the calls to use writeRead() -- or wrapped them the same way.

IAmNickNack:
Is there any particular reason these couldn't be default methods on the interface rather than the base class>

stefanhaustein:
> Is there any particular reason these couldn't be default methods on the interface rather than the base class>

I2CRegisterDataReader / I2CRegisterDataWriter (where they are declared) doesn't have a reference to I2C.

I2C is an interface and IIRC interfaces can't override methods to provide a default implementation(?)

stefanhaustein:
I have closed this for now because there wasn't any progress for days and I don't want to create another fork for some more urgent stuff I'd like to do. The suggested additions are preserved here: https://github.com/Pi4J/pi4j/issues/522

Issue #514, opened 2025-10-01
V4-Snapshot: Issues with ST7789  (pi5)


stefanhaustein:
On a PI5 with autocontext (=gpiod)
 
- When writing to a gpio pin from a timer task, I get the exception below
- Elsewhere, write calls just seems to do nothing 
- Reads seem to work fine

Reproduction:

- Connect a st7789 display to the pins used in the test
- Run https://github.com/Pi4J/pi4j-drivers/blob/main/src/test/java/com/pi4j/drivers/display/graphics/st7789/St7789DriverTest.java
- Observe that (apart from noting working) not even the backlight turns on, which is controlled by a separate pin. I suspect "nothing renders" is caused by the data/command select and rst pins don't work -- I have checked separately that spi works)

Some extra context can be found on https://github.com/Pi4J/pi4j/issues/500, which was focussing primarily on a related but different problem and was just closed.

```
Exception in thread "Timer-0" com.pi4j.io.exception.IOException: Failed to set state for output DOUT-25 to LOW
	at com.pi4j.plugin.gpiod.provider.gpio.digital.GpioDDigitalOutput.state(GpioDDigitalOutput.java:97)
	at com.pi4j.io.gpio.digital.DigitalOutputBase.off(DigitalOutputBase.java:276)
	at com.pi4j.io.gpio.digital.DigitalOutputBase.off(DigitalOutputBase.java:44)
	at com.pi4j.drivers.display.graphics.st7789.St7789Driver.command(St7789Driver.java:110)
	at com.pi4j.drivers.display.graphics.st7789.St7789Driver.command(St7789Driver.java:116)
	at com.pi4j.drivers.display.graphics.st7789.St7789Driver.setPixels(St7789Driver.java:159)
	at com.pi4j.drivers.display.graphics.GraphicsDisplay$DriverEntry.transferBuffer(GraphicsDisplay.java:330)
	at com.pi4j.drivers.display.graphics.GraphicsDisplay$DriverEntry.transferBuffer(GraphicsDisplay.java:277)
	at com.pi4j.drivers.display.graphics.GraphicsDisplay.transferBuffer(GraphicsDisplay.java:252)
	at com.pi4j.drivers.display.graphics.GraphicsDisplay.flush(GraphicsDisplay.java:113)
	at com.pi4j.drivers.display.graphics.GraphicsDisplay$1.run(GraphicsDisplay.java:236)
	at java.base/java.util.TimerThread.mainLoop(Timer.java:572)
[...]

PR #525, opened 2025-10-08
SPI writes of length greater than the kernel allows are broken on mul…

taartspi:
…tiple spi_ioc_transfer records and all written to the spi device on a single IOCTL  

Code function properly with data buffers that fit into one spi_ioc_transfer       Buffers that require multiple spi_ioc_transfer in a single ioctl fails error 90   msg size.  Using the debugger  and forcing the creation of two spi_ioc_transfer.  Their sizes are 100 and 92, correct for my data.   I reviewed the structure definition for data type and JNA definition.  I suspect something is not correct in this regard and the second entry is not properly mapped the C interface 
as the first entry has incorrect padding or some similar problem 

Stefan offered to review the code for this error.

### DIFF
--- a/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
+++ b/plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
@@ -41,9 +41,7 @@ import com.sun.jna.Structure;
 import com.sun.jna.ptr.IntByReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import java.io.File;
-import java.util.Arrays;
 import java.util.Objects;
 import java.util.Scanner;
 
@@ -337,22 +335,17 @@ public class LinuxFsSpi extends SpiBase implements Spi {
      * write
      *
      *   SPI_BUFFSIZ most often is set to 4096.  See initialize()
-     * This implementation can write blocks greater than 4096 byte
+     *   for the current configured size.
+     * This implementation can write blocks greater than SPI_BUFFSIZ byte
      * 'however', read and understand how this is accomplished.
      *
-     * A write of data no greater than 4096 bytes is accomplished with
+     * A write of data greater than SPI_BUFFSIZ bytes is accomplished with
      * a single SPI write operation.
      * So CE line low, write bytes, CE line high
      *
-     * A write greater than 4096 bytes will be segmented to multiple writes
-     * each 4096  bytes in length, the last write is the MOD value.
-     * So, CE line low, write first 4096 bytes CE line high.  This
-     * pattern will repeat until the last bytes length MOD 4096 are written.
-     * This means multiple SPI transaction with you SPI device. If CE line
-     * toggling creates problems with your SPI device your application
-     * can use a vacant GPIO configured as an Output pin and your
-     * application keep the CE pin low during the duration of the call to
-     * spi,write.
+     * A write greater than SPI_BUFFSIZ bytes will be segmented to multiple
+     * spi_ioc_transfer.  All spi_ioc_transfer will be written to the kernel
+     * in a single IOCTL
      *
      *
      * @param data data array of bytes to be written
@@ -364,33 +357,62 @@ public class LinuxFsSpi extends SpiBase implements Spi {
     public int write(byte[] data, int offset, int length) {
         Objects.checkFromIndexSize(offset, length, data.length);
 
+        int fullEntries = (length < SPI_BUFFSIZ) ? 1 : length / SPI_BUFFSIZ;
+        int partialEntries = (((fullEntries * SPI_BUFFSIZ) < length) ? 1 : 0);
+        int numberXferEntries = fullEntries + partialEntries;
+        spi_ioc_transfer[] transferArray = getTransferArray(numberXferEntries);
+        Memory[] peerArray = getPeerMem(numberXferEntries, SPI_BUFFSIZ) ;
         int start = offset;
-        while (start < data.length) {
-            PeerAccessibleMemory buf = new PeerAccessibleMemory(SPI_BUFFSIZ);
-            spi_ioc_transfer txEntry = new spi_ioc_transfer() ;
-            int end = Math.min(data.length, start + SPI_BUFFSIZ);
-            byte[] chunk = Arrays.copyOfRange(data, start, end);
-            buf.write(0, chunk, 0 , chunk.length);
+        int lastOffset = data.length + offset ;
+        int entryNum = 0;
+        while (start < lastOffset) {
+            //  PeerAccessibleMemory buf = new PeerAccessibleMemory(SPI_BUFFSIZ);
+            spi_ioc_transfer txEntry = transferArray[entryNum];
+            int end = Math.min(lastOffset, start + SPI_BUFFSIZ);
+            peerArray[entryNum].write(0, data, start, end - start);
             // set fields in transfer msg
-            txEntry.tx_buf = buf.getPeer();
+            txEntry.tx_buf = Memory.nativeValue(peerArray[entryNum]) ;
             txEntry.rx_buf = 0;
             txEntry.bits_per_word = BITS8;
             txEntry.speed_hz = config.baud();
             txEntry.delay_usecs = 0;
-            txEntry.len = chunk.length ;
-            txEntry.cs_change = 1 ;
-            int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(1), txEntry);
-            logger.trace("[SPI::WRITE] <- Number bytes {} ", ret);
-            if (ret < 0) {
-                logger.error("Could not write SPI message. ret {}, error: {}", ret, Native.getLastError());
-                length = 0;
-            }
+            txEntry.len = end - start;
+            txEntry.cs_change = 0;
+            transferArray[entryNum] = txEntry;
+            entryNum++;
             start += SPI_BUFFSIZ;
+        }
+        // last entry has chip select change  set to  deassert
+        transferArray[entryNum - 1].cs_change = 1;
+        int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(numberXferEntries), transferArray);
+        logger.trace("[SPI::WRITE] <- Number bytes {} ", ret);
+
+        if (ret < 0) {
+            logger.error("Could not write SPI message. ret {}, error: {}", ret, Native.getLastError());
+            length = 0;
         }
 
         return length;
     }
 
+    private spi_ioc_transfer[] getTransferArray(int numberEntries) {
+        spi_ioc_transfer[] transferArray = (spi_ioc_transfer[]) new spi_ioc_transfer().toArray(numberEntries);
+
+        for (int i = 0 ; i <  numberEntries ; i++){
+            transferArray[i] = new spi_ioc_transfer();
+            PeerAccessibleMemory buf = new PeerAccessibleMemory(SPI_BUFFSIZ) ;
+            transferArray[i].tx_buf = buf.getPeer() ;
+        }
+        return transferArray ;
+    }
+    private static  Memory[] getPeerMem(int numberEntries, int size){
+        Memory[] memArray = new Memory[numberEntries] ;;
+        for (int i = 0 ; i <  numberEntries ; i++){
+            PeerAccessibleMemory buf = new PeerAccessibleMemory(size) ;
+            memArray[i] = new Memory(size) ;
+        }
+        return memArray ;
+    }
     /**
      * Extension of Memory to allow direct access to the native peer pointer. This is required because
      * the {@link spi_ioc_transfer} structure uses a long for the tx_buf and rx_buf fields but a


plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -364,33 +357,62 @@ public int write(byte b) {
     public int write(byte[] data, int offset, int length) {
         Objects.checkFromIndexSize(offset, length, data.length);
 
+        int fullEntries = (length < SPI_BUFFSIZ) ? 1 : length / SPI_BUFFSIZ;
```
stefanhaustein: If length < SPI_BUFSIZ, there is no full entry and the next line will cover this case without the ternary here

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -364,33 +357,62 @@ public int write(byte b) {
...
-                length = 0;
-            }
+            txEntry.len = end - start;
+            txEntry.cs_change = 0;
+            transferArray[entryNum] = txEntry;
+            entryNum++;
             start += SPI_BUFFSIZ;
         }
+        // last entry has chip select change  set to  deassert
+        transferArray[entryNum - 1].cs_change = 1;
```
stefanhaustein: As this is a significant change in behaviour compared to the state before, I think we should isolate this in a separate PR and carefully check with existing examples / drivers and consider options for enabling this:

On one hand, existing code will be working around the 4096 limit using multiple physical transfers to represent a single logical transfer, and without changes there, will continue to do so, unaware of this change here, potentially being broken now.

On the other hand, doing this might allow us to get the BMP 280 working without an extra pin.

IMHO, the best way to control (and test) this behaviour might be a configuration option.

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -364,33 +357,62 @@ public int write(byte b) {
...
+
+        if (ret < 0) {
+            logger.error("Could not write SPI message. ret {}, error: {}", ret, Native.getLastError());
+            length = 0;
+        }
 
         return length;
     }
 
+    private spi_ioc_transfer[] getTransferArray(int numberEntries) {
```
stefanhaustein: could this be static?

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -364,33 +357,62 @@ public int write(byte b) {
...
+            entryNum++;
             start += SPI_BUFFSIZ;
         }
+        // last entry has chip select change  set to  deassert
+        transferArray[entryNum - 1].cs_change = 1;
+        int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(numberXferEntries), transferArray);
+        logger.trace("[SPI::WRITE] <- Number bytes {} ", ret);
+
+        if (ret < 0) {
+            logger.error("Could not write SPI message. ret {}, error: {}", ret, Native.getLastError());
```
stefanhaustein, line 391: I think we should discuss consistency across plugins / providers for this case separately:

Java only does this when running into a blocked read case -- errors are always reported as exceptions. 

And there the return value will never be 0

Moreover, writing 0 bytes could be a degenerated case where just nothing is written, and user code would need to be aware of this and distinguish the cases...

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -364,33 +357,62 @@ public int write(byte b) {
...
+            length = 0;
+        }
 
         return length;
     }
 
+    private spi_ioc_transfer[] getTransferArray(int numberEntries) {
+        spi_ioc_transfer[] transferArray = (spi_ioc_transfer[]) new spi_ioc_transfer().toArray(numberEntries);
```
stefanhaustein: Doesn't this allocate a continuous memory array....

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -364,33 +357,62 @@ public int write(byte b) {
     public int write(byte[] data, int offset, int length) {
         Objects.checkFromIndexSize(offset, length, data.length);
 
+        int fullEntries = (length < SPI_BUFFSIZ) ? 1 : length / SPI_BUFFSIZ;
+        int partialEntries = (((fullEntries * SPI_BUFFSIZ) < length) ? 1 : 0);
+        int numberXferEntries = fullEntries + partialEntries;
```
stefanhaustein: This can be written without the two helper variables as

int numberXferEntries = (length + SPI_BUFSIZ -1) / SPI_BUFSIZ;

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -364,33 +357,62 @@ public int write(byte b) {
...
+        }
 
         return length;
     }
 
+    private spi_ioc_transfer[] getTransferArray(int numberEntries) {
+        spi_ioc_transfer[] transferArray = (spi_ioc_transfer[]) new spi_ioc_transfer().toArray(numberEntries);
+
+        for (int i = 0 ; i <  numberEntries ; i++){
+            transferArray[i] = new spi_ioc_transfer();
```
stefanhaustein: ... which you are overwriting with new "segmented" parts here?

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -364,33 +357,62 @@ public int write(byte b) {
...
-            }
+            txEntry.len = end - start;
+            txEntry.cs_change = 0;
+            transferArray[entryNum] = txEntry;
+            entryNum++;
             start += SPI_BUFFSIZ;
         }
+        // last entry has chip select change  set to  deassert
+        transferArray[entryNum - 1].cs_change = 1;
+        int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(numberXferEntries), transferArray);
```
stefanhaustein: Not sure -- I think you might need to hand over transferArray[0] here

stefanhaustein:
I suspect the main problem might be as follows (mostly guessing, as I am not really familiar with jna):

- new spi_ioc_transfer().toArray(numberEntries) allocates the array entries in continuous memory, i.e. like a traditional c array. The java array elements point to the entries there
- These are then overwritten with new, non-continuous individual structs in the loop below
- The ioctl call might need to use the first array entry instead of the array

EDIT: If the last bullet was right, the code should not work for the one element case, so I think it's fine to hand in the array...

Generally, I'd prefer if we'd first fix the loop code -- and then add this improvement as a second step, as the code is quite complex already and it's good to have a "safe" point for reference, reducing the amount / complexity of diffs.... That said, if the main problem can be solved here quickly, I guess it's fine to take it from there....

stefanhaustein:
Hm, I think the ioctl call should construct the native array correctly from the java array no matter what (but not 100% sure, probably worth trying a continuous array anyway). So another issue might be that the struct size is wrong. This wouldn't matter much for a single call, in particular if it's to big, but for the array call it would shift the address of the next entry... 

EDIT: The size seems to be correct though as far as I can tell...

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: I think it is the last entry that needs the CS_CHANGE set.   When I get this running I will watch the line and verify.   I have an IKA Logic  logic analyzer.   Easy to use on common bus'

taartspi:
Thanks Stefan.  I will investigate/redo the array questions your pose.    I

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
stefanhaustein: I intended to refer to this line: 
```
 int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(numberXferEntries), transferArray);
```
But if this would be a problem, the one-element case wouldn't work, so I no longer think it is. It won't hurt to try though I guess...

I think we should move back to not setting CS_CHANGE until we know whether this is causing #514, but this is a separate issue.

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: yea, i think so..     After change same error.  Setting max size to 48 to create 4 entries, if i set the call to ioctl as though only one entry it runs and sets some leds.  Any number of entries fails.  So second entry is bad. Looking the array when calling the ioctl the padding that was added is still visible. 
   spi_ioc_transfer[] transferArray = new spi_ioc_transfer[numberEntries];
        //(spi_ioc_transfer[]) new spi_ioc_transfer().toArray(numberEntries);
        for (int t = 0 ; t < numberEntries ; t++) {
            transferArray[t] = new spi_ioc_transfer();
        }

stefanhaustein: Your comment looks like the opposite, so let's double-check we don't have some kind of misunderstanding...

What I was trying to suggest is this:

Keep this line -- it will allocate continuous memory:
```
spi_ioc_transfer[] transferArray = (spi_ioc_transfer[]) new spi_ioc_transfer().toArray(numberEntries);
```

Remove the "for t" loop below -- it will overwrite the continuous entries with newly allocated ones in diverging memory locations
```
// for (int t = 0 ; t < numberEntries ; t++) {
//    transferArray[t] = new spi_ioc_transfer();
// }
```
Change this line 
```
 int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(numberXferEntries), transferArray);
``` 
 to
```
 int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(numberXferEntries), transferArray[0]);
```

Not sure about this -- it's just what I would try...

taartspi: yes  the for loop setting another instance was wrong.   I got to that point yesterday afternoon, sorry I didn't update tthen to save you thinking more about this, I wanted to do more testing prior to saying anything.     I will have time later today to validate the question on CS operating correctly.  appreciate the help.

taartspi:
I made updates recommended by Stefan's review and fixed the actual bug with his help/  In testing the CS line assert did not function as i expected while using  a logic analyzer. . I will include what to me seems conflicting information. Searching seems the cs_change flag creates a great number of questions. @DigitalSmile you have some ffm code using this flag, appreciate your involvement and  thought on making this change in FFM.    The actual kernel docs are not very clear on defining the flag  

AI Overview
How to use it: When you need to send multiple SPI messages back-to-back without raising the CS line between them, you pass an array of spi_ioc_transfer structures to a
 single SPI_IOC_MESSAGE ioctl call. The cs_change flag in each of the transfers controls the CS line.
cs_change = 0: Keeps the CS line asserted (active) after the transfer is complete. This is used for a series of transfers that should be treated as a single transaction.
cs_change = 1: Deasserts the CS line after the transfer is complete. This forces a chip-select toggle and signals the end of a transaction. 


AI Overview
The cs_change is a member of the spi_ioc_transfer structure used with the SPI_IOC_MESSAGE ioctl command 
for SPI devices to control the Chip Select (CS) line. Setting cs_change to 1 will keep the CS line active
 (de-asserted) between SPI transfers, while 0 will de-assert it after the transfer is complete. 
 You can also break up a multi-byte transfer into single-byte transfers, each with its own 
 cs_change setting, to have finer control over when the CS line is toggled, 
 as shown in NVIDIA Developer Forums and GitHub.

DigitalSmile:
> @DigitalSmile you have some ffm code using this flag, appreciate your involvement and thought on making this change in FFM. The actual kernel docs are not very clear on defining the flag

As far as I understand the kernel docs and some researching (https://git.zx2c4.com/linux-rng/commit/drivers/spi?h=jd/vdso-test-harness&id=c00d5e93ea018786d98670fc1d0dab4c36c2217c and https://stackoverflow.com/questions/57430453/how-to-set-chip-select-in-spi-programming-for-multipleslaves-struct-spi-ioc-t), in Linux you are opening an spi device with CS preconfigured. E.g. opening `/dev/spidev0.1` essentially means `open device bus 0 with chip select 1`. And during the work with this device changing `cs_change` does not make any sense to kernel driver, as it is managed internally. So by the end of the day changing CS effect depends on the driver you use.

But I don't feel it is a provider thing to split buffer for user if it exceeds `SPI_BUFFSIZ `. I believe it is supposed to be managed by user driver code, we can just add necessary checks in place to indicate problems. Let's keep provider code more atomic and as simple as possible.

stefanhaustein:
> I made updates recommended by Stefan's review

IMHO The most important part here is making sure cs_change is 0, as the current state at head presumably causes https://github.com/Pi4J/pi4j/issues/514 

Existing code is breaking up a large transfer into chunks clientside to work around the 4096 limit. I assume that is not something we want to break.

> and fixed the actual bug with his help/ In testing the CS line assert did not function as i expected while using a logic analyzer. . I will include what to me seems conflicting information. Searching seems the cs_change flag creates a great number of questions. @DigitalSmile you have some ffm code using this flag, appreciate your involvement and thought on making this change in FFM. The actual kernel docs are not very clear on defining the flag

The description of the flag seems quite clear to me:

- Each transfer will set CS in any case, as selecting the device (and deselecting others) is required for the transfer 
- If the flag is set, CS will be released after the transfer
- If the flag is set on any *intermediate* transfer, this will implicitly result in a toggle. I suspect the pulse width will depend on the usec parameter

Tom, you mentioned that checking with a logic analyzer gave you unexpected results. What behaviour did you see there?

- Supporting the cs_change flag would be nice, as it would presumably allow us to support BMP280 without an extra pin
- It will be a change breaking existing clientside chunking logic, so l I think it will require a dedicated configuration option.

taartspi:
CS_CHANGE zero all the time.   
If 1 in in an entry that is not the last, CS remains low and is raised when the next entry begins processing.  If the last entry CS remains LOW until the next transfer happens.  
After knowing how it works I can now understand their explanation and fill in the words they left out.
@DigitalSmile you mentioned being an atomic operation, as all transfer take place within the same ioctil it is atomic,  Are still against this being within ffm providers ?

stefanhaustein:
> If 1 in in an entry that is not the last, CS remains low and is raised when the next entry begins processing. If the last entry CS remains LOW until the next transfer happens. After knowing how it works I can now understand their explanation and fill in the words they left out. 

Could it be that low and high are mixed up in your description (as confusingly high means inactive for CS)? 

Here is how I understand the spec:
        
``` 
                      Entry 1, cs_change=0    Entry 2, cs_change=0
high / inactive   _
                   \__________________________________________________ ........
low / active             

                      Entry 1, cs_change=0    Entry 2, cs_change=1
high / inactive   _                                                  ______________.....
                   \________________________________________________/ 
low / active             

                      Entry 1, cs_change=1    Entry 2, cs_change=1
high / inactive   _                         _                         ______________.....
                   \_______________________/ \_______________________/ 
low / active             
```

Is this consistent with your measurement, or did you see something else?

> @DigitalSmile you mentioned being an atomic operation, as all transfer take place within the same ioctil it is atomic, Are still against this being within ffm providers ?

As far as I understand Pi4j, it's a user friendly api making some abstractions. There is no way for a pi4j user to re-create an atomic multipart transfer like this PR does. 

If pi4j would be a low-level API, exposing limitations to users, I'd expect the IOCTL call to be exposed to users directly.

taartspi:
Rather than words to describe the CS line, pictures
[spi_logic.pdf](https://github.com/user-attachments/files/22876181/spi_logic.pdf)
  
Ignore the MISO traffic. There is a mcp3008 also connected

stefanhaustein:
> Rather than words to describe the CS line, pictures [spi_logic.pdf](https://github.com/user-attachments/files/22876181/spi_logic.pdf)
> 
> Ignore the MISO traffic. There is a mcp3008 also connected

Was this with cs_change=0 everywhere? Might make sense to keep track of this in the PDF, in case we need to get back to this?

So this looks like there is still a gap -- even with the multi-chunk call -- and all this work was basically for nothing? 

(I must admit I wasn't really concerned about the gap as the clock pulse should keep normal stuff working BUT when (ab)using SPI for led matrix displays, this is probably an issue... :-/  Seems to limit us to 4096 / 12 = 341 leds if we can write two bit per byte and we need 24 bit per led....)

taartspi:
Yes. Always cs = 0.  Yes the clocks stop and start when the spi buffer is too small  I too think chips should tolerate this.  
Yes the matrix only configures some leds when i force this to happen with a matrix of 8 leds so it is unusable. As far as matrix using that ws281 I don't think i am worried about that. The web site already says this device is difficult to use on the pi spi. And recommends a stand alone device to offload the spi work.   And i can delete that example neopixel and update with a pointer to the web site details.   
all that said.  A few weeks ago i asked two questions about the spi. 
A. Do nothing tell folks to increase the buffer size and live with it. 
B. Try and complete the ioctl work.  
No one really said a great deal on my questions

taartspi:
I will find time tomorrow to test a bmp280 spi with a small buffer to see if it creates problems

stefanhaustein:
> Yes. Always cs = 0. Yes the clocks stop and start when the spi buffer is too small I too think chips should tolerate this. Yes the matrix only configures some leds when i force this to happen with a matrix of 8 leds so it is unusable. As far as matrix using that ws281 I don't think i am worried about that. The web site already says this device is difficult to use on the pi spi. And recommends a stand alone device to offload the spi work. And i can delete that example neopixel and update with a pointer to the web site details. 

I don't think a limit of > 100 LEDs is too bad. Might make sense to just point out what the limits are to this approach. People trying to use a 8x8 matrix will still find it useful. 

> all that said. A few weeks ago i asked two questions about the spi. A. Do nothing tell folks to increase the buffer size and live with it. B. Try and complete the ioctl work. No one really said a great deal on my questions

In my opinion, the current state at head is good (option C?) :

- It will automatically split larger requests, avoiding exposing somewhat surprising limits on the user
- As your measurement has shown,  this PR does not fundamentally improve the situation, but adds complexity and increases dynamic memory allocations, so I think we shouldn't proceed with it -- although it's sad that the work is wasted

There is something else I wanted to talk about:

- I don't think we should change the cs behaviour in general, as existing code will expect cs_change=0
- That said, I wonder if setting cs_change=1 for the last chunk would allow us to drive a BMP/BME 280 in SPI mode with a "standard" CS pin -- instead of controlling a user-driven dedicated CS-pin "manually". If this is the case, I think setting cs_change for the last part of SPI transfers should be part of the SPI configuration builder.

taartspi:
Been thinking about this while getting bmp280 spi connected. Thinking if clock interruption are usable with the bmp it might be  another chip would fail and debug be difficult for the user.  Another reason to jettison this code.   

Looked at my bmp spi code, forgot i handled the CS line directly which is your reason for a CS option and why u asked for spi config with no chip select value required. CS control easy, haven't tried spi-0cs to see if no cs device tree objects are easy to omit or another investigation 

I could add a long detail to the docs about buffer size and our behavior.  Give some example math on what will be written in a single ioctl and where to look on increasing the kernel spi buffer size.  Frank can suggest how much i add the java docs vs a link to the web site and place the explanation on the web pages

taartspi:
The use of multiple spi_ioc_transfer to segment a buffer larger than the OS allows is not to be done.   It does work, but the pause in clocking data as one spi_ioc_transfer is completed means a ws281 LED controller considers the data stream completed and the next spi_ioc_transfer is the beginning of a completely  new data stream.     If an application must send very large data buffers the OS /sys/module/spidev/parameters/bufsiz should be researched and configured larger

PR #531, opened 2025-10-19
#496 Fix for permissions to be applied by udev

### LINKED ISSUE #496: FFM plugin: PWM sometimes throws an exception after creation
```
[main] INFO com.pi4j.runtime.impl.DefaultRuntime - Pi4J context/runtime successfully initialized.
PWM config create
Error: Error during call to method 'open' with data '[/sys/class/pwm/pwmchip0/pwm2/enable, 2]': Permission denied (13)
```

Possible cause: _slow creation of needed files in initialisation, so udev rules are not applied in time_

DigitalSmile:


DigitalSmile:
@FDelporte can you please check the fix from this PR and confirm that the issues is gone?

FDelporte (approved):

PR #537, opened 2025-10-25
Fix RDWR Memory Layouts

IAmNickNack:
The current implementation maps these fields to JAVA_INT and defines the buffer as a statically allocated array. However, the i2c message struct is defined in C as:

```C
struct i2c_msg {
    __u16 addr;
    __u16 flags;
    __u16 len;
    __u8 *buf;
};
```

The current implementation also defines `msgs` as a statically allocated array, but the RDRW data struct is defined in C as:

```C
struct i2c_rdwr_ioctl_data {
     struct i2c_msg *msgs;
     __u32 nmsgs;
};

```

### DIFF
--- a/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/common/i2c/rdwr/I2CMessage.java
+++ b/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/common/i2c/rdwr/I2CMessage.java
@@ -2,10 +2,11 @@ package com.pi4j.plugin.ffm.common.i2c.rdwr;
 
 import com.pi4j.plugin.ffm.common.Pi4JLayout;
 
+import java.lang.foreign.Arena;
 import java.lang.foreign.MemoryLayout;
 import java.lang.foreign.MemorySegment;
+import java.lang.foreign.SegmentAllocator;
 import java.lang.foreign.ValueLayout;
-import java.lang.invoke.MethodHandle;
 import java.lang.invoke.VarHandle;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


eitch (approved):

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/common/i2c/rdwr/I2CMessage.java
```diff
@@ -98,23 +102,23 @@ public I2CMessage from(MemorySegment buffer) throws Throwable {
...
-        var bufMemorySegment = invokeExact(MH_BUF, buffer);
-        var buf = new byte[len];
-        for (int i = 0; i < buf.length; i++) {
-            buf[i] = bufMemorySegment.getAtIndex(ValueLayout.JAVA_BYTE, i);
-        }
-        return new I2CMessage(address, flags, len, buf);
+
+        var bufferSegment = (MemorySegment)VH_BUFFER.get(buffer, 0L);
+        var bytes = new byte[len];
+        bufferSegment.reinterpret(len).asByteBuffer().get(bytes, 0, len);
```
DigitalSmile: @IAmNickNack In general it is not safe to get `ByteBuffer` from `MemorySegment` - it is not guaranteed to be backed (see https://download.java.net/java/early_access/loom/docs/api/java.base/java/lang/foreign/MemorySegment.html#asByteBuffer() in exception description).

The other thing, is that you are creating an intermediate object in heap, it is better to stick with `MemorySegment` methods in this classes.
Can you please fix this?

PR #538, opened 2025-10-26
Spi writeRead   write and read operation contained in single ioctl

taartspi:
Addition to the SPI interface, supported in linuxfs but not pigpio, and not in ffm as of yet.

Allows SPI devices that require a write to setup the chip and then the following read.  If there is a required delay between the write and read or after the read and before the CS is released they are each available as parameter.

Review, the many space additions were not seen at the user level in gitk. Appears the git push initiated a cleanup step with different rules.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/spi/Spi.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/spi/Spi.java
@@ -368,21 +368,19 @@ public interface Spi extends IO<Spi, SpiConfig, SpiProvider>, AutoCloseable, IOD
     }
 
     // ------------------------------------------------------------------------------------
-    //  writeRead
+    //  writeThenRead
     // ------------------------------------------------------------------------------------
     /**
      * This function writes bytes to the SPI device and then reads bytes from the device
      * Write data is taken from the 'buffer' byte array, data
      * read back from the SPI device is then copied to the 'read' byte array
      *
-     * @param buffer the array of bytes to write to the SPI device
+     * @param write the array of bytes to write to the SPI device
 
      * @param read   Buffer to contain read data
-     * @return Returns 0 if OK, otherwise PI_BAD_HANDLE, PI_BAD_SPI_COUNT, or PI_SPI_XFER_FAILED.
-     * @see "http://abyz.me.uk/rpi/pigpio/cif.html#spiWrite"
      */
-    default int writeRead(byte[] buffer,  byte[] read) {
-        return writeRead(buffer, 0, buffer.length, (short) 0, read, 0, read.length, (short) 0);
+    default void  writeThenRead(byte[] write,  byte[] read) {
+         writeThenRead(write, 0, write.length, (short) 0, read, 0, read.length, (short) 0);
     }
 
     /**
@@ -390,16 +388,14 @@ public interface Spi extends IO<Spi, SpiConfig, SpiProvider>, AutoCloseable, IOD
      * Write data is taken from the 'buffer' byte array, data
      * read back from the SPI device is then copied to the 'read' byte array
      *
-     * @param buffer the array of bytes to write to the SPI device
+     * @param write the array of bytes to write to the SPI device
      * @param writeLength
      *               Number bytes written from buffer to the SPI
      * @param read   Buffer to contain read data
      * @param readlength the number of bytes to read (read &amp; read))
-     * @return Returns 0 if OK, otherwise PI_BAD_HANDLE, PI_BAD_SPI_COUNT, or PI_SPI_XFER_FAILED.
-     * @see "http://abyz.me.uk/rpi/pigpio/cif.html#spiWrite"
      */
-    default int writeRead(byte[] buffer, int writeLength,  byte[] read,  int readlength) {
-        return writeRead(buffer, 0, writeLength, (short) 0, read, 0, readlength, (short) 0);
+    default void writeThenRead(byte[] write, int writeLength,  byte[] read,  int readlength) {
+        writeThenRead(write, 0, writeLength, (short) 0, read, 0, readlength, (short) 0);
     }
 
 
@@ -421,19 +417,17 @@ public interface Spi extends IO<Spi, SpiConfig, SpiProvider>, AutoCloseable, IOD
      *                           start writing to the SPI device from and the position
      *                           used as the starting offset position to place data bytes
      *                           read back from the SPI device.
-     * @param writeNumberOfBytes Number of bytes written from write buffer
+     * @param writeLength        Number of bytes written from write buffer
      * @param writeDelayUsecs    Delay after SPI write record processed by kernel before the
      *                           read SPI record is processed. Value in usecs.
      * @param read               Buffer to contain read data
      * @param readOffset         Offset within the read buffer to begin placing read data
-     * @param readNumberOfBytes  the number of bytes to transfer/exchange (read &amp; read))
-     * @param readDelayUsecs    Delay after SPI read record processed by kernel before the
+     * @param readLength         The number of bytes to transfer/exchange (read &amp; read))
+     * @param readDelayUsecs     Delay after SPI read record processed by kernel before the
      *                           read SPI record is processed. Value in usecs.
-     * @return Returns 0 if OK, otherwise PI_BAD_HANDLE, PI_BAD_SPI_COUNT, or PI_SPI_XFER_FAILED.
-     * @see "http://abyz.me.uk/rpi/pigpio/cif.html#spiWrite"
      */
-    default int writeRead(byte[] write, int writeOffset, int writeNumberOfBytes,  short writeDelayUsecs, byte[] read, int readOffset, int readNumberOfBytes,  short readDelayUsecs) {
-        throw new IllegalStateException("writeRead Not supported in this provider. \n See https://www.pi4j.com/documentation/providers/");
+    default void writeThenRead(byte[] write, int writeOffset, int writeLength,  short writeDelayUsecs, byte[] read, int readOffset, int readLength,  short readDelayUsecs) {
+        throw new IllegalStateException("writeThenRead Not supported in this provider. \n See https://www.pi4j.com/documentation/providers/");
     }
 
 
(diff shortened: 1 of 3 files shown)


stefanhaustein (commented):
Does this resolve the BMP 280 extra pin problem?

I2C has writeRead() methods, why not use names consistent with this?

pi4j-core/src/main/java/com/pi4j/io/spi/Spi.java
```diff
@@ -366,4 +366,59 @@ default int transfer(ByteBuffer buffer, int offset, int length) {
...
+     *
+     * @param buffer the array of bytes to write to the SPI device
+     * @param writeLength
+     *               Number bytes written from buffer to the SPI
+     * @param read   Buffer to contain read data
+     * @param readlength the number of bytes to read (read &amp; read))
+     * @return Returns 0 if OK, otherwise PI_BAD_HANDLE, PI_BAD_SPI_COUNT, or PI_SPI_XFER_FAILED.
+     * @see "http://abyz.me.uk/rpi/pigpio/cif.html#spiWrite"
+     */
+    default int registerRead(byte[] buffer, int writeLength,  byte[] read,  int readlength) {
```
DigitalSmile: I guess there is already a method with similar signature `transfer` (I am using it now with ffm, but for different purpose). May be we should stick with it?

https://github.com/Pi4J/pi4j/blob/develop/pi4j-core/src/main/java/com/pi4j/io/spi/Spi.java#L128

DigitalSmile (changes requested):

pi4j-core/src/main/java/com/pi4j/io/spi/Spi.java
taartspi: I believe transfer expects the returned data miso interleaved with the mosi.   The bmp spec says. Write the data and then read it back

DigitalSmile: That's essentially what it does, from javadoc:

```
     * This function transfers (writes/reads simultaneously) multiple bytes with the SPI
     * device associated with the handle.  Write data is taken from the 'write' byte array
     * from the given 'writeOffset' index to the specified length ('numberOfBytes').  Data
     * read from the SPI device is then copied to the 'read' byte array at the given 'readOffset'
     * using the same length ('numberOfBytes').  Both the 'write' and 'read' byte arrays must
     * be at least the size of the defined 'numberOfBytes' + their corresponding offsets.
```

taartspi:
Stefan. Yes this removes the manual CS gpio in those sensors. Earlier in this PR i listed the BMP example device PR that uses this new code.   
Yes on the name change

stefanhaustein:
> Yes this removes the manual CS gpio in those sensors. Earlier in this PR i listed the BMP example device PR that uses this new code.

That's awesome!

I wonder if we should add a new small interface for this, shared between spi and i2c -- this would greatly simplify drivers for chips supporting both protocols. I think everything else (IODataReader, IODataWriter, register access) could be implement in Java on top of a single native writeRead call...

pi4j-core/src/main/java/com/pi4j/io/spi/Spi.java
```diff
@@ -366,4 +366,60 @@ default int transfer(ByteBuffer buffer, int offset, int length) {
...
+     *
+     * @param buffer the array of bytes to write to the SPI device
+     * @param writeLength
+     *               Number bytes written from buffer to the SPI
+     * @param read   Buffer to contain read data
+     * @param readlength the number of bytes to read (read &amp; read))
+     * @return Returns 0 if OK, otherwise PI_BAD_HANDLE, PI_BAD_SPI_COUNT, or PI_SPI_XFER_FAILED.
+     * @see "http://abyz.me.uk/rpi/pigpio/cif.html#spiWrite"
+     */
+    default int writeRead(byte[] buffer, int writeLength,  byte[] read,  int readlength) {
```
stefanhaustein: Why have this convenience method (opposed to just two byte arrays)? I think JRE typically only has "as few parameters as possible" and  the full version?

pi4j-core/src/main/java/com/pi4j/io/spi/Spi.java
```diff
@@ -366,4 +366,60 @@ default int transfer(ByteBuffer buffer, int offset, int length) {
...
+     *                           read SPI record is processed. Value in usecs.
+     * @param read               Buffer to contain read data
+     * @param readOffset         Offset within the read buffer to begin placing read data
+     * @param readNumberOfBytes  the number of bytes to transfer/exchange (read &amp; read))
+     * @param readDelayUsecs    Delay after SPI read record processed by kernel before the
+     *                           read SPI record is processed. Value in usecs.
+     * @return Returns 0 if OK, otherwise PI_BAD_HANDLE, PI_BAD_SPI_COUNT, or PI_SPI_XFER_FAILED.
+     * @see "http://abyz.me.uk/rpi/pigpio/cif.html#spiWrite"
+     */
+    default int writeRead(byte[] write, int writeOffset, int writeNumberOfBytes,  short writeDelayUsecs, byte[] read, int readOffset, int readNumberOfBytes,  short readDelayUsecs) {
```
stefanhaustein: This has the offset and count parameters swapped compared to the I2C writeRead method (sorry noticed this only recently).

Perhaps we can introduce a new method name here (e.g. writeThenRead(?)) and then back-port this to I2C and deprecate the one with the surprising parameter order there?

pi4j-core/src/main/java/com/pi4j/io/spi/Spi.java
taartspi: The array length could vary and the write and read array can be different in length

stefanhaustein: Sorry for being ambiguous... I was suggesting to have

```
 default int writeRead(byte[] buffer,  byte[] read) {
```

Matching what JRE typically does wrt. convenience methods (= both extremes, as little params and all params, but nothing in the middle)

taartspi: added simpler method signature

stefanhaustein: Thanks!

Sorry for being "piecemeal" here... Would you mind also renaming to "writeThenRead", so we can have the same method in I2C without a conflict with the existing method, which has the parameters swapped from the "standared" order...

pi4j-core/src/main/java/com/pi4j/io/spi/Spi.java
taartspi: I edited the javadoc. To explain this code is a write and then a read.    

True that at the HW level they are both transfers where only the tx or rx buffer are used but a detail of no value here. 
Also added a simpler implementation

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -393,6 +370,75 @@ public int write(byte[] data, int offset, int length) {
...
+        rxEntry.speed_hz = config.baud();
+        rxEntry.delay_usecs = readDelayUsec;
+        rxEntry.cs_change = 0;
+        rxEntry.len = readNumberOfBytes;
+
+        int ret = libc.ioctl(fd, SPI_IOC_MESSAGE(totalRecords), (Object) transferArray);
+
+        logger.trace("[SPI::RegRead] <- Number bytes read {} ", ret);
+        if (ret < 0) {
+            logger.error("Failed ioctl SPI message. ret {}, error: {}", ret, Native.getLastError());
```
stefanhaustein: Can we make this throw?

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
```diff
@@ -393,6 +370,75 @@ public int write(byte[] data, int offset, int length) {
...
+        }
+        rxBuf.read(0, read, readOffset, readNumberOfBytes);
+
+        return readNumberOfBytes;
```
stefanhaustein: As this is just the input parameter, can we make the method void?

Returning a value here might mislead users to try to put this into a loop, as one would do with stream read operations....

taartspi:
The simple name change to writeThenRead is in my local repo. I was reviewing for return value and the error documentation and will have time later today to complete.   Looks like i will pull docs from the interface to this linuxfs spi to remove some details pertinent to pigpio.  Exception throw.  I think on the new signature we can  consider such changes and i will give it thought. But any consideration for those changes in the older interface details should wait to the ffm implementation review since we agreed users might have code impacts in the V4 move

plugins/pi4j-plugin-linuxfs/src/main/java/com/pi4j/plugin/linuxfs/provider/spi/LinuxFsSpi.java
taartspi: @stefanhaustein @DigitalSmile @eitch @FDelporte    We decided to use writeThenRead, simple,  changed here, still need to add an additional  signature in the I2C provider to complement its registerRead.  I pushed this code for Stefan's discussion on using throws and return type void.   I don't think we should make this change in existing code but this one newly add signature in linuxfs could make this change and the ffm code is open to any of these changes.  This is is here for comments to decide what we want to do ?????

taartspi:
@stefanhaustein @DigitalSmile @eitch    This completes the discussion on return void and throwing an exception for error/    I made this change in linuxfs SPI code in the new signature writeThenRead.  In the case of SPI both write and read are a single atomic operation so the spi chipselect can be used with a bmp/bme sensor, a separate GPIO for chipselect is no longer required   
Also in i2c interface deprecated writeRead and make a duplicate behavior called  writeThenRead for naming consistency

eitch (approved):

eitch:
It looks ok, i think we can merge it

taartspi:
Don't merge plz.   I let the tool do the last merge and i need to verify what it did

taartspi:
@FDelporte   Reviewing the merge I saw a stale method name in a trace message.  Fixed it.

Ready to merge

PR #541, opened 2025-10-29
#540 Fix DIRECT implementation of I2C write

### LINKED ISSUE #540: FFM I2C implementation DIRECT   writeRegister executes, but correct data update did not happen
writeRegister(byte[] register, byte[] data, int length)

@DigitalSmile .
The writeRegister does not operate correctly.  No errors, but the intended storage is not updated.

Linuxfs start w bytes stop

ffm start w byte byte restart w bytes stop using FFM internalWrite
The ffm code uses 2 entries in the ioctl parms array rather than a single entry with the tx_buf containing the register bytes and the data bytes, 
If you need i2c traces or other detail let me know.

DigitalSmile:


eitch (approved):

taartspi (commented):
Thanks Nick

PR #542, opened 2025-10-29
#535 Set default to I2CDirect for provider, added logging, fix Segmen…

### LINKED ISSUE #535: FFM I2C  SMBus protocol does not support reading multiregister devices.
Regression testing with a AT24C512 EEPROM.     Test attempts to read data pointed at by  a two byte register address.  
[main] TRACE com.pi4j.devices.at24c512.AT24C512 - >>> enter: readEEPROM register 0342  numByte : 2
Exception in thread "main" java.lang.UnsupportedOperationException: SMBus protocol does not support reading multiregister devices. Please, use I2CDirect or I2CFile provider instead.
	at com.pi4j.plugin.ffm@4.0.0-SNAPSHOT/com.pi4j.plugin.ffm.providers.i2c.impl.I2CSMBus.readRegister(I2CSMBus.java:121)
	at com.pi4j@4.0.0-SNAPSHOT/com.pi4j.io.i2c.I2CRegisterDataReader.readRegister(I2CRegisterDataReader.java:139)
	at com.pi4j@4.0.0-SNAPSHOT/com.pi4j.io.i2c.I2CRegisterDataReader.readRegister(I2CRegisterDataReader.java:171)
	at com.pi4j.devices/com.pi4j.devices.at24c512.AT24C512.readEEPROM(AT24C512.java:228)
	at com.pi4j.devices/com.pi4j.devices.at24c512.AT24C512_App.main(AT24C512_App.java:208)
     

This is a device that requires a REPEATED RESTART.  I will include info from linuxfs and the datasheet
.  
Random Read
A random read begins in the same way as a byte write operation does to load in a new data word address. This is
known as a “dummy write” sequence; however, the data byte and the Stop condition of the byte write must be omitted
to prevent the part from entering an internal write cycle. Once the device address and word address are clocked in
and acknowledged by the EEPROM, the bus host must generate another Start condition. The bus host now initiates
a current address read by sending a Start condition, followed by a valid device address byte with the R/W bit set to
logic ‘1’. The EEPROM will ACK the device address and serially clock out the data word on the SDA line. All types
of read operations will be terminated if the bus host does not respond with an ACK (it NACKs) during the ninth clock
cycle. After the NACK response, the host may send a Stop condition to complete the protocol, or it can send a Start
condition to begin the next sequence.

Image

Image

DigitalSmile:
…tAllocator in I2CMessage and RDWRData

### DIFF
--- a/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/I2CFFMProviderImpl.java
+++ b/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/I2CFFMProviderImpl.java
@@ -29,10 +29,12 @@ public class I2CFFMProviderImpl extends I2CProviderBase implements I2CProvider {
     public I2C create(I2CConfig config) {
         var bus = new I2CBusFFM(config);
 
-        var functions = bus.getFunctionalityMap();
-        logger.debug("{} - bus I2C functions support:", bus.getBusName());
-        for (var entry : functions.entrySet()) {
-            logger.debug("\t{} - {}", entry.getKey(), entry.getValue());
+        if (logger.isDebugEnabled()) {
+            var functions = bus.getFunctionalityMap();
+            logger.debug("{} - bus I2C functions support:", bus.getBusName());
+            for (var entry : functions.entrySet()) {
+                logger.debug("\t{} - {}", entry.getKey(), entry.getValue());
+            }
         }
 
         var impl = config.i2cImplementation();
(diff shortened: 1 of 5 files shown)


plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/I2CFFMProviderImpl.java
```diff
@@ -26,16 +29,26 @@ public int getPriority() {
     public I2C create(I2CConfig config) {
         var bus = new I2CBusFFM(config);
 
+        var functions = bus.getFunctionalityMap();
```
eitch: Perhaps put this in an if debug is enabled

eitch (changes requested):

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/I2CFFMProviderImpl.java
eitch: Sorry, i meant put the foreach block in an if statement checking if debug/trace is enabled

FDelporte (approved):

FDelporte:
Requested changes by @eitch were applied

eitch (approved):

PR #543, opened 2025-10-29
Refactor IO parameters

FDelporte:
As discussed, with V4 we may introduce some breaking changes regarding the IO parameters to remove confusion between pin, address, channel, etc.

I tried to find a first starting point for this change in this pull request. I'm sure it's not finished, e.g. some tests fail at this moment. But **it's a starting point** for further discussion and cooperation to check if this is something we want to bring in V4.0.0 or wait for later...


pi4j-core/src/main/java/com/pi4j/io/gpio/GpioConfig.java
```diff
@@ -25,22 +25,21 @@
...
+ * @param <CONFIG_TYPE>
  * @author Robert Savage (<a href="http://www.savagehomeautomation.com">http://www.savagehomeautomation.com</a>)
  * @version $Id: $Id
- * @param <CONFIG_TYPE>
  */
 public interface GpioConfig<CONFIG_TYPE extends Config>
-        extends AddressConfig<CONFIG_TYPE>, IOConfig<CONFIG_TYPE> {
+    extends PinConfig<CONFIG_TYPE>, IOConfig<CONFIG_TYPE> {
 
     String BUS_NUMBER = "bus-number";
```
taartspi: I will need to see the usage to understand the name

pi4j-core/src/main/java/com/pi4j/io/pwm/Pwm.java
```diff
@@ -60,17 +60,17 @@ static PwmConfigBuilder newConfigBuilder(Context context){
...
     /**
      * Get the GPIO pin number/address of this PWM instance.
      *
      * @return gpio pin number/address
      */
-    default int address(){
-        return getAddress();
+    default int getAddress() {
+        return config().getAddress();
     }
```
taartspi, line 74: I expected address would become channel ???

taartspi (commented):
Bus.  I think it is the pwmChipX. The vale of X.  Makes sense 
  
Also in pwm. We have address. I expected this would be channel. 

How do we organize unique instances. Before it was address by IO-TYPE.  What does this map(s) look like now ?

pi4j-core/src/main/java/com/pi4j/io/gpio/GpioConfig.java
FDelporte: Unique instances is indeed a problem. I already improved the Registry to be able to make the distinction between different IO Types, but need to further investigate as the integration test is still failing.

DigitalSmile:
May I suggest not to rename parameters, but to add new ones and add `@Deprecate` to legacy? I guess the breaking change is better to be done after removal of all other providers.

FDelporte:
> May I suggest not to rename parameters, but to add new ones and add `@Deprecate` to legacy? 

I tried but because the way the configs inherit from a nested structure of interfaces, this becomes pretty hard...

FDelporte:
@DigitalSmile I further revised this proposal, trying to find a good balance between new parameters, renaming, and deprecating `address`. Almost all flags are green now, except a test in the FFM plugin. Any idea what's going wrong?

See build https://github.com/Pi4J/pi4j/actions/runs/18976808907/job/54198562656

```
Error:    I2CTest.testCreation:60 » IOAlreadyExists IO instance with address 28 already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.
Error:    I2CTest.testReadDirect:181 » IOAlreadyExists IO instance with address 28 already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.
Error:    I2CTest.testReadFile:240 » IOAlreadyExists IO instance with address 28 already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.
Error:    I2CTest.testReadSMBus:111 » IOAlreadyExists IO instance with address 28 already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.
Error:    I2CTest.testWriteDirect:144 » IOAlreadyExists IO instance with address 28 already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.
Error:    I2CTest.testWriteFile:214 » IOAlreadyExists IO instance with address 28 already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.
Error:    I2CTest.testWriteSMBus:84 » IOAlreadyExists IO instance with address 28 already exists in the Pi4J runtime context; unable to create a new instance using this reserved id.
```

DigitalSmile:
> @DigitalSmile I further revised this proposal, trying to find a good balance between new parameters, renaming, and deprecating `address`. Almost all flags are green now, except a test in the FFM plugin. Any idea what's going wrong?
> 
> See build https://github.com/Pi4J/pi4j/actions/runs/18976808907/job/54198562656
> 

I have fixed the issue. Changed `device()` to `bus()`, so that you cannot create same buses within pi4j.

DigitalSmile:
@FDelporte It's better looking now! One small thing to improve is to add some javadoc explanation to deprecated methods what to use instead. E.g.

```
    /**
     * @deprecated use {@link #pin(Integer pin)} instead
     * <p>address.</p>
     *
     * @param address a {@link java.lang.Integer} object.
     * @return a BUILDER_TYPE object.
     */
    @Deprecated(forRemoval = true)
    BUILDER_TYPE address(Integer address);
```

FDelporte:
More @deprecated JavaDoc was added.
Ready for re-review @DigitalSmile @eitch ...

libraries/pi4j-library-pigpio/src/main/java/com/pi4j/library/pigpio/PiGpioPacket.java
```diff
@@ -40,18 +40,18 @@
...
- *      uint32_t p1;
- *      uint32_t p2;
- *      union
- *      {
- *          uint32_t p3;
- *          uint32_t ext_len;
- *          uint32_t res;
- *      };
- *  } cmdCmd_t;
+ * typedef struct
```
eitch: Perhaps we can fix this whitespace?

pi4j-core/src/main/java/com/pi4j/io/gpio/analog/AnalogOutputProvider.java
```diff
@@ -40,58 +40,58 @@ public interface AnalogOutputProvider extends AnalogProvider<AnalogOutputProvide
     /**
      * <p>create.</p>
      *
-     * @param address a {@link java.lang.Integer} object.
+     * @param pin a {@link java.lang.Integer} object.
      * @param <T> a T object.
      * @return a T object.
      */
-    default <T extends AnalogOutput> T create(Integer address) {
+    default <T extends AnalogOutput> T create(Integer pin) {
```
eitch: If we are talking about pin, do we now mean the hardware pin number and not BCM address? E.g. Pin 15 here: https://pinout.xyz/pinout/pin15_gpio22/

pi4j-core/src/main/java/com/pi4j/platform/Platform.java
```diff
@@ -94,95 +96,112 @@ default <T extends Provider> T provider(Class<T> providerClass) throws ProviderN
...
                 // iterate over providers looking for a matching class/interface
-                for(Provider provider : providers().values()) {
+                for (Provider provider : providers().values()) {
                     if (providerClass.isInstance(provider)) {
                         return (T) provider;
                     }
                 }
             }
-        } catch (ClassNotFoundException e){}
+        } catch (ClassNotFoundException e) {
```
eitch, line 149: I know this is not part of the PR, but an empty catch? That isn't a good thing, now is it?

pi4j-test/src/main/java/com/pi4j/test/Main.java
```diff
@@ -90,20 +89,20 @@ public static void main(String[] args) {
 
         input.addListener((AnalogValueChangeListener) event -> {
             logger.info("ANALOG INPUT [");
-            logger.info(String.valueOf(event.source().address()));
+//            logger.info(String.valueOf(event.source().address()));
```
eitch: Why are we commenting out these log statements? Either remove them, or change to trace/debug?

eitch (commented):
I like the changes, and think this is the right direction, but i'm wondering if pin is the right idea. Am i wrong, that till now address for GPIO was the BCM number, and with pin we are now changing to the hardware pin? Shouldn't we then at least offer both schemes? Looking at https://pinout.xyz/pinout/pin15_gpio22/ i would say pin means the hardware address, and thus we should be able to also use BCM then, as otherwise all our users will need to change their addresses which will probably lead to quite a bit of frustration.

pi4j-core/src/main/java/com/pi4j/io/gpio/analog/AnalogOutputProvider.java
FDelporte: Good remark, actually we use the BCM numbering in all examples. So maybe instead of `pin` it should be `bcm`? That would definitely take away even more confusion about how to define digital in and out...

eitch: True, bcm might be better, but then again we should think about offering a mechanism where users can use physical pin numbering too.

FDelporte: Not sure what the best solution is...

Current: `Integer pin();`

Possible alternatives:

1/ One setter method with a record
```
record Pin(NumberingScheme ns, Integer value)
enum NumberingScheme { BCM, PHYSICAL, ... } 
 
Pin pin();
```

2/ Private values set by two different methods (later more, e.g. WiringPi would have been needed in earlier cases)
```
private NumberingScheme schema
private value

Integer bcm(int value) {
   schema = NumberingScheme.BCM;
   value = value;
}
Integer physical(int value) {
   schema = NumberingScheme.PHYSICAL;
   value = value;
}
```

What do you think?

eitch: I was first thinking to not use the numbering scheme internally, but only on setting the pin, but perhaps approach two would be better. Or a combination: for existing code it would be easier to just replace the .address(int) call with .bcm(int), but if someone wants to use the physical, then we can use the first idea.

Thus we can properly abstract away the numbering when new versions arise.

pi4j-test/src/main/java/com/pi4j/test/Main.java
FDelporte: cleaned up

pi4j-core/src/main/java/com/pi4j/io/gpio/analog/AnalogOutputProvider.java
FDelporte: Because we only use BCM numbering at this moment, I think the best solution (for now) is to use BcmConfig instead of PinConfig. That will make it as clear as possible. I did this change and added two helper methods, in case someone needs to convert from one value to another:

```
BoardInfoHelper.getBcmFromPhysical(int value)
BoardInfoHelper.getBcmFromWiringPi(int value)
```

If we need a way to use physical pin numbers, or any other type later, we can still extend with extra Config-classes.

eitch:
wonderful

eitch (approved):

PR #544, opened 2025-11-03
Revert deprecated Serial implementations

FDelporte:
As the FFM plugin will provide serial implementation, I propose to revert the earlier deprecations related to any serial functionality.

This was done in 2025-03-24 - V3.0.1
Related to discussion https://github.com/Pi4J/pi4j/discussions/308


eitch (approved):
I agree, with FFM we have a good reason to keep Serial in pi4j.

PR #546, opened 2025-11-12
Correctly identify IOs in the registry based on bcm and/or bus and/or device and/or channel

FDelporte:
Current implementation doesn't except multiple I2C devices on the same bus.


taartspi:
Working from my phone which limits investigation. But i think spi will have the same problem

FDelporte:
@taartspi good remark! I made this a uniform approach across all IO Types with a new method `getUniqueIdentifier`. Unit test got extended to validate this.

stefanhaustein (approved):
The comment / formatting changes make it hard to spot actual functional changes, perhaps that's something that could be done separately for the whole code base as a preparation (or followup) step (if it's done by a tool?)

That said, looks like the right fix to me, so lgtm :)

PR #547, opened 2025-11-12
Add deprecated address to DigitalConfigBuilder

FDelporte:
To avoid too much breaking changes, as requested by @stefanhaustein

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/gpio/digital/impl/DefaultDigitalOutputConfigBuilder.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/gpio/digital/impl/DefaultDigitalOutputConfigBuilder.java
@@ -63,7 +63,7 @@ public class DefaultDigitalOutputConfigBuilder
      */
     @Override
     public DigitalOutputConfigBuilder address(Integer address) {
-        this.properties.put(GpioConfig.ADDRESS_KEY, String.valueOf(address));
+        this.properties.put(GpioConfig.BUS_KEY, String.valueOf(address));
         return this;
     }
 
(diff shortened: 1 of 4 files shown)


stefanhaustein (commented):
Doesn't this need to be wired up somehow / somewhere to map to bcm?

Could that perhaps be in the builder only, so that we don't need to bring it back to DigitalConfig?

FDelporte:
@stefanhaustein good thinking, should be better now...

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/impl/DefaultDigitalOutputConfigBuilder.java
```diff
@@ -63,7 +63,7 @@ public static DigitalOutputConfigBuilder newInstance(Context context) {
      */
     @Override
     public DigitalOutputConfigBuilder address(Integer address) {
-        this.properties.put(GpioConfig.ADDRESS_KEY, String.valueOf(address));
```
stefanhaustein, line 66: This looks odd... Is this intended? If so, why / what does this do?

FDelporte: all part of the existing builder code, so following the same approach to set the BUS value.

stefanhaustein: Wouldn't it be more clear to just remove this and just fall back to the default from DigitalConfigBuilder? Is BUS the same as bcm?

FDelporte: Oh that's an error, fixed now, must be BCM_KEY

stefanhaustein: Still, why not just remove the method? 
Won't the changed default interface method still end up doing the same? 
Less code, fewer opportunity for errors...?

stefanhaustein (approved):

PR #550, opened 2025-11-13
Change pulse so it always returns to the inverse of the activation state

gyorik:
On DigitalOutput, the original implementation toggles the activation on pulse.
This is problematic especially when using the async version of the function.
If two requests collide, it can produce  a state where the original state of the output is inverted when two pulse calls are made.

e.g.
async pulse(...);
async pulse(...);

Original state is LOW
First call changes state to HIGH
First call timeout
Second call changes state to HIGH
Second call timeout
First call changes state to LOW (inverts current)
Second call inverts state so output is HIGH

imo this is should not be the expected behavior


eitch (approved):
I agree with your PR, thanks for this fix.

Issue #551, opened 2025-11-14
Bring clear instructions and parameters for PWM usage


FDelporte:
It's probably me, but I find PWM the hardest IO Type to use... It's unclear how to map GPIO pin numbers to chip/channel parameters for hardware PWM, how to configure on different types of RPi's, etc.

Current documentation: https://www.pi4j.com/documentation/io-types/pwm/

I guess this can be fixed by extending the documentation and parameters in the following way:

## Hardware PWM

Only a limited number of hardware-accelerated PWMs are available. These can be checked with:

```
$ ls -l /sys/class/pwm/
pwmchip0

$ pinctrl | grep PWM
18: a3    pd | lo // GPIO18 = PWM0_CHAN2
19: a3    pd | lo // GPIO19 = PWM0_CHAN3
45: op dh pd | hi // FAN_PWM/GPIO45 = output
```

### Implementation

* [x] Parameters should be `chip` and `channel`
* [x] The plugin PWM provider must throw an error if `chip` and `channel` are not provided with `PwmType.HARDWARE`
* [x] The plugin PWM provider should log a warning if a `bcm` is provided with `PwmType.HARDWARE`
* [x] Implementation changes needed in the plugins to allow this usage
* [ ] @DigitalSmile can we add com.pi4j.plugin.ffm.providers.pwm.PwmFFMSoftware ?

```
# Chip can be retrieved with a helper method
# This returns 0 for non-RP1 (pre-RPi 5) and correct number for RP1 (RPi 5)
var chip = PwmChipUtil.getPWMChip();

# Example based on the output above for `GPIO18 = PWM0_CHAN2`
var pwmConfig = Pwm.newConfigBuilder(pi4j)
    .pwmType(PwmType.HARDWARE)
    .chip(chip)
    .channel(2)
    .build();
```

### Documentation

* [x] Describe how to enable, configure, and use HARDWARE PWM with chip and channel on different RPi models
* [x] Create JBang example
* [x] Document what happens if you try to use hardware PWM on a non-hardware PWM pin
* [x] Document frequency/duty cycle usage and limitations 

## Software PWM

Software PWM is implemented entirely in code, and it can generate PWM signals on any pin that supports general-purpose output. But it has less accurate timing due to CPU load and OS scheduling.

### Implementation

* [x] Parameters should be `bcm`
* [x] The plugin PWM provider throw an error if i`bcm` is not provided with `PwmType.SOFTWARE`
[...]

PR #553, opened 2025-11-14
Improve PWM parameters and clear difference between SOFTWARE/HARDWARE definitions

FDelporte:
Improvements based on #551

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/pwm/impl/DefaultPwmConfig.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/pwm/impl/DefaultPwmConfig.java
@@ -48,8 +48,9 @@ public class DefaultPwmConfig
     extends IOConfigBase<PwmConfig>
     implements PwmConfig {
 
-    protected Integer bus = null;
+    protected Integer chip = null;
     protected Integer channel = null;
+    protected Integer bcm = null;
 
     // private configuration properties
     protected Integer dutyCycle = null;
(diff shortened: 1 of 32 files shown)


pi4j-core/src/main/java/com/pi4j/io/pwm/impl/DefaultPwmConfig.java
```diff
@@ -48,8 +48,9 @@ public class DefaultPwmConfig
     extends IOConfigBase<PwmConfig>
     implements PwmConfig {
 
-    protected Integer bus = null;
+    protected Integer chip = null;
     protected Integer channel = null;
+    protected Integer bcm = null;
```
IAmNickNack, line 53: Is `bcm` necessary for PWM? I don't think it's relevant for hardware PWM, at least.

taartspi (approved):
Looks good.  I will find the PR for documentation of this change

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/pwm/PwmFFMProviderImpl.java
```diff
@@ -24,6 +22,21 @@ public int getPriority() {
...
     @Override
     public Pwm create(PwmConfig config) {
+        // validate PWM type
+        if (config.pwmType() != PwmType.HARDWARE) {
+            throw new IOException("The FFM PWM provider only supports HARDWARE PWM");
+        }
+
+        // validate the config
+        if (config.chip() == null || config.channel() == null) {
+            throw new IllegalArgumentException("PWM Chip and Channel are needed for hardware PWM with the LinuxFS I/O provider");
```
eitch: Here you write LinuxFS I/O, but we're in FFM, which is correct?

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/pwm/PwmFFMProviderImpl.java
```diff
@@ -24,6 +22,21 @@ public int getPriority() {
...
+        }
+
+        // Warn for unneeded config
+        if (config.pwmType() == PwmType.HARDWARE && config.bcm() != null) {
+            logger.warn("You specified a BCM value for the PWM, but this is not needed for hardware PWM. Please specify chip and channel instead.");
```
eitch, line 37: Do we want to warn, or throw? Since users need to refactor anyhow, perhaps simply throw? Or what argues against?

eitch (commented):
I have two comments, otherwise looks good.

Issue #554, opened 2025-11-15
PWM device can't be removed in v4 SNAPSHOT


IAmNickNack:
I can create a PWM device. 

Creation logs successfully: 

```
2025-11-15 09:21:48.912 INFO  [DefaultDispatcher-worker-1] [i.g.i.p.g.s.s.D.Pwm                 ] Creating device with config: {channel=0, bus=2, duty-cycle=50, frequency=220}
```

This config creates a device in `/sys/class/pwm/pwmchip0/` and I can use it as expected.

Removing the same device throws an error: 

```
2025-11-15 09:21:54.661 INFO  [DefaultDispatcher-worker-2] [i.g.i.p.g.s.s.D.Pwm                 ] Removing device with id: PWM-0
2025-11-15 09:21:54.667 ERROR [DefaultDispatcher-worker-2] [i.g.i.p.g.s.s.D.Pwm                 ] Failed to remove device: Error during call to method 'open' with data '[/sys/class/pwm/pwmchip2/unexport, 1]': No such file or directory (2)
com.pi4j.exception.Pi4JException: Error during call to method 'open' with data '[/sys/class/pwm/pwmchip2/unexport, 1]': No such file or directory (2)
	at com.pi4j.plugin.ffm.common.file.FileDescriptorNative.open(FileDescriptorNative.java:36)
	at com.pi4j.plugin.ffm.providers.pwm.PwmFFMHardware.shutdownInternal(PwmFFMHardware.java:165)
	at com.pi4j.plugin.ffm.providers.pwm.PwmFFMHardware.shutdownInternal(PwmFFMHardware.java:18)
	at com.pi4j.registry.impl.DefaultRuntimeRegistry.remove(DefaultRuntimeRegistry.java:162)
```

There are a couple of issues here:

`unexport` for this device is actually located at `/sys/class/pwm/pwmchip0/unexport`:

```
--w--w---- 1 root gpio 16384 Nov  8 17:27 /sys/class/pwm/pwmchip0/unexport
```

This code: https://github.com/Pi4J/pi4j/blob/develop/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/pwm/PwmFFMHardware.java#L165 appears to be looking in the wrong place. 

```java
var exportFd = file.open(CHIP_PATH + bus + CHIP_UNEXPORT_PATH, FileFlag.O_WRONLY);
```

Should this be?:

```java
[...]

PR #555, opened 2025-11-15
Use correct path for unexport

IAmNickNack:
This PR intends to resolve an issue described by issue https://github.com/Pi4J/pi4j/issues/554

--- 

The problem appears to be due to confusion over the terms `bus` and `channel` and brings the `unexport` inline with their usage in the constructor: https://github.com/Pi4J/pi4j/blob/75df994e4a18c65c2994563392cf7a350f51fea5/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/pwm/PwmFFMHardware.java#L49

I understand that there are efforts elsewhere to address PWM confusions. This PR simply solves the issue for the current nomenclature. The long-term fix will no doubt be provided in response to https://github.com/Pi4J/pi4j/issues/551

### DIFF
--- a/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/pwm/PwmFFMHardware.java
+++ b/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/pwm/PwmFFMHardware.java
@@ -162,8 +162,8 @@ public class PwmFFMHardware extends PwmBase implements Pwm {
             return super.shutdownInternal(context);
         }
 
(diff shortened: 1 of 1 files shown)


IAmNickNack:
I did attempt to provide unit tests for this change, but this seemed to require changes to `FileDescriptorNativeMock` which I was apprehensive to make.

FDelporte:
If we can merge #553, then this won't be needed, as it already contains a clear fix with chip versus channel

```
        var exportFd = file.open(CHIP_PATH + chip + CHIP_UNEXPORT_PATH, FileFlag.O_WRONLY);
        file.write(exportFd, getByteContent(channel));
        file.close(exportFd);
```

FDelporte:
Handled in #553

PR #558, opened 2025-11-17
Smoke test two

taartspi:
Uses junit
SPI test works using linuxfs,  fffm needs to implement writeThenRead, I spoke Friday with Ncik. He was going to review recent PRs and help out with ffm.

Digittal output works with with gpiod, fails with ffm.  I will debug more to gather info and then open an issue.

Serial test not done.  I pasted some test code to change and use for this test, but nothing to review now.


pi4j-test/src/main/java/com/pi4j/test/README.md'
```diff
@@ -0,0 +1,64 @@
+
```
stefanhaustein, line 1: The file name seems to have a single quote at the end

pi4j-test/src/test/java/com/pi4j/test/smokeTest/SmokeTestControl.java
```diff
@@ -0,0 +1,114 @@
...
+import com.pi4j.plugin.ffm.providers.serial.SerialFFMProviderImpl;
+import com.pi4j.util.Console;
+
+import org.junit.jupiter.api.*;
+
+
+
+
+
+public class SmokeTestControl {
```
stefanhaustein: As this class has the actual test methods, I'd name this one "SmokeTest".

I am not sure about the other class. I find myself constantly scrolling back and forth to understand what's going on. 

Typically, tests are pretty self-contained.

So I'd lean towards just inlining it here for now.

If this becomes too much code in one file, I'd split into separate classes per type, "PwmSmokeTest", "SerialSmokeTest" etc.

I'd also lean towards limiting the scope of variables as much as possible. Can it be a local variable? Make it a local variable. Keeping stuff local provides assurance that it can't have an influence on other things going on elsewhere.

That said, having any smoke tests is a big step forward, so it's probably better to go ahead and check stuff. We can improve things any time.  Would be nice to name the class containing the @Test methods "...Test" though.

pi4j-test/src/test/java/com/pi4j/test/smokeTest/SmokeTest.java
```diff
@@ -0,0 +1,315 @@
...
+import com.pi4j.util.Console;
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+
+import com.pi4j.util.DeferredDelay.* ;
+
+
+public class SmokeTest {
+    static final String I2C_PROVIDER = "ffm-i2c"; //"linuxfs-i2c";
```
stefanhaustein: One way to organize this without manual changes would be as follows:

- Have an abstract base class "AbstractGpioSmokeTest" with a ctor that takes the provider and has all the gpio test methods
- For each provider, have a subclass "FfmGpioSmokeTest", "GpiodGpioSmokeTest" that just calls the super constructor with the provider but has no other methods

Another way might be parameterized tests.

taartspi:
@stefanhaustein    Moved code into a single file.  also added a directory provider so the next category of smoke test can have its own directory.  Each test has its constants within the test,  be easy to break these into individual classes.   Been thinking about your comment on whether to use small classes, not sure on doing it.      The spi tests works linuxfs fails ffm as the writeThenRead needs implementation.   I need to complete the serial test

pi4j-test/src/test/java/com/pi4j/test/smokeTest/provider/ProviderSmokeTest.java
```diff
@@ -0,0 +1,264 @@
...
+
+        DigitalInput gpio25InMonitor = createDigitalInput(PIN_GPIO25);
+        // Validate monitor is LOW, test control state HIGH
+        if (gpio25InMonitor.state() == DigitalState.LOW) {
+            gpio24OutTest.high();
+        }
+        try {
+            Thread.sleep(1000);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
```
taartspi: The sleep is here while i debug ffm. Without the pause the test usually fails. When i have enough details i will open an issue to s needed

stefanhaustein:
Looks good but please remove the trailing single quote in the file name of README.md

pi4j-test/src/main/java/com/pi4j/test/README.md'
stefanhaustein: This seems to be unresolved.

pi4j-test/src/test/java/com/pi4j/test/smoketest/provider/ProviderSmokeTest.java
```diff
@@ -0,0 +1,264 @@
...
+    @Test
+    public void testSpi() {
+        final int idValueMskBMP = 0x58;   // expected chpId value BMP280
+        final int idValueMskBME = 0x60;   // expected chpId value BME280
+
+        Spi spi = createSPIDevice();
+        final int chipId = 0xD0;
+        try {
+            Thread.sleep(100);
+        } catch (InterruptedException e) {
```
stefanhaustein, line 112: Just let the test throw, Junit will take care...

pi4j-test/src/test/java/com/pi4j/test/smokeTest/provider/ProviderSmokeTest.java
```diff
@@ -0,0 +1,264 @@
...
+        Pwm pwm = createPWM(PWM_CHANNEL);
+        pwm.on(50, 1) ;
+        Thread.sleep(10000);  // wait 10 seconds while listener counts flashes
+
+        Assertions.assertTrue(pwmFlashes == 10, "TestPWM flashing 10 times in 10 seconds");
+    }
+
+    @Test
+    public void testGpioIn() {
+          final int PIN_GPIO16 = 16;
```
stefanhaustein: indent

pi4j-test/src/test/java/com/pi4j/test/smokeTest/provider/ProviderSmokeTest.java
```diff
@@ -0,0 +1,264 @@
...
+            gpio26OutControl.high();
+        }
+        DigitalState state = gpio16InTest.state();
+
+        Assertions.assertTrue(state  ==  DigitalState.HIGH, "TestGpioIn  now HIGHs");
+    }
+
+    @Test
+    public void testGpioOut() {
+        final int PIN_GPIO24 = 24;
```
stefanhaustein: If there is no "semantic" name for the constant and it's only used once, perhaps it would be more straightforward to inline it?

pi4j-test/src/test/java/com/pi4j/test/smokeTest/provider/ProviderSmokeTest.java
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    @Test
+    public void testSerial() {
+        Assertions.assertTrue(false, "TestSerial expected read on loopback");
+    }
+
+
+    private I2C  createI2cBMPDevice() {
+        final String I2C_PROVIDER = "ffm-i2c";
```
stefanhaustein: Sorry for causing confusion. Configuration constants should probably remain static final and at the top of the file. 

My concern about keeping stuff local was mostly about input/output instances etc.

stefanhaustein (approved):
Fine to go ahead as-is, but please fix the README file name.

I have added some optional nits to prove that I actually looked at the code :)

stefanhaustein:
Do you have any guess what's going wrong with the CI?

eitch:
I think this was just a fluke. Retrying

eitch (changes requested):
There seem to be some broken tests, and missing dependencies:

https://github.com/Pi4J/pi4j/actions/runs/19525975870/job/55933065003#step:4:7104

taartspi:
The test fails.  The serial test has not been completed. My work.  The spi fail has an issue against ffm to implement the writeThenRead    The gpioOut only fails on ffm and i am getting details to explain this in an issue.  I can put a sleep in the gpioOut test and it seems the thread sleeping lets some ffm code run properly     I could close the PR and Stefan use my branch to comment but the PR makes that an easier process    I don't see the dependency error run directly.  I will fix that

taartspi:
I don’t know why it runs in my workspace but CI requires added changes.
And. I need see how i set this up so it doesn’t run in CI.

taartspi:
@eitch @stefanhaustein    I was using ffm in my local, seems the CI does not,  and my attempt to flip to linuxfs failed and I will not make any more attempts tonight.   I pushed this to make sure it is closer to what we want ??

If yes, I plan to use a utility to switch between what providers are used,  and complete the Serial test,   And figure out the behavior difference in testing output pin in linuxfs vs ffm.

eitch:
Hey @taartspi it was just a case of duplicate dependencies. The first had scope test, the second without and this is conflicting. I hope it's ok, that pushed a commit to fix these issues.

eitch:
@taartspi ok, i finally understand the problem. If i had looked at the entire logs, then i would have seen it sooner:

Loading FFM plugin does a `PermissionHelper.checkUserPermissions(this)`, which throws an exception, thus the FFM providers don't get added to the runtime. But this is only logged as a warning, so the test continues to execute.

Since your test explicitly tries to get the FFM provider, it fails as the provider is not loaded.

On a Raspberry Pi we don't have this issues, as on the Raspberry Pi the checks don't fail. On the CI they fail as well as when we run it locally on a non Raspberry Pi system.

To fix the issue i disabled the test if it is not running on a Raspberry Pi. Now the build is good here on the CI.

eitch (approved):

PR #559, opened 2025-11-17
Fixing an error.

mores:
In comments from https://github.com/Pi4J/pi4j/pull/547/files

> Oh that's an error, fixed now, must be BCM_KEY


stefanhaustein:
Thanks! Would you mind trying if removing the method altogether also fixes the problem (the default method should handle the mapping already)?

FDelporte:
Strange, thought I changed it already. Thanks for pull request @mores !

PR #560, opened 2025-11-18
PWM now uses chip and channel.  Made code change so the path is  /sys…

taartspi:
…/class/pwm/pwmchipCHIP/CHANNEL   At present the provider ffm integration test for PWM fails.  Attempting to run the test exits with no detail and ignoring the debugger.  Will dig more and then ask Nick.  I think the test path it uses needs update.

With this change the example now works
  ```java     
  private static Pwm createPWM(int channel) {
        var chip = PwmChipUtil.getPWMChip();
        var configPwm = Pwm.newConfigBuilder(pi4j)
            .channel(channel)
            .pwmType(PwmType.HARDWARE)
            .provider(PWM_PROVIDER)     // ffm-pwm
            .initial(50)
            .frequency(1)
            .chip(chip)   
            .shutdown(0) 
            .build();
        pwm = pi4j.create(configPwm);
        return pwm;
    }
```

### DIFF
--- a/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/pwm/PwmFFMHardware.java
+++ b/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/pwm/PwmFFMHardware.java
@@ -46,7 +46,7 @@ public class PwmFFMHardware extends PwmBase implements Pwm {
         super(provider, config);
         this.chip = config.chip();
         this.channel = config.channel();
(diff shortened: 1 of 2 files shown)


taartspi:
plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/unit/PWMTest.java
This file needs updates to correct chip channel usage.  Doing so did not fix the test failure so more will be needed

FDelporte (approved):

FDelporte:
Unit test is fixed. Thanks @taartspi for the changes

PR #561, opened 2025-11-26
Smoke test implementation

taartspi:
…thod writeThenRead.  gpio test fails on ffm need debug.    Serial test needs implementation

It appears the pom dependency ffm is not required

This is the program I want used for smokeTest so comments welcome.   I need for implement serial test for ffm and debug the ffm fails in one test.


taartspi:
Fails for module-info not finding the provider ffm.. ??  need to investigate

pi4j-test/src/main/java/com/pi4j/test/Main.java
```diff
@@ -35,10 +35,23 @@
 import com.pi4j.io.spi.SpiBus;
 import com.pi4j.io.spi.SpiChipSelect;
 import com.pi4j.io.spi.SpiMode;
+//mport com.pi4j.plugin.linuxfs.internal.LinuxGpio;
```
eitch: can you clean up this import?

pi4j-test/src/main/java/com/pi4j/test/Main.java
```diff
@@ -388,40 +436,46 @@ private static class ProviderContext{
...
 
-        private String getProviderName(){
-            return useProviders ;
+        private String getProviderGroupName() { return group ;}
+
+        private Context getContext() {
+            return pi4j;
         }
+
         private String i2cName = "";
```
stefanhaustein: I think i2cProviderName (etc.) would be more clear.

pi4j-test/src/main/java/com/pi4j/test/Main.java
```diff
@@ -388,40 +436,46 @@ private static class ProviderContext{
...
         private String spiName = "";
         private String pwmName = "";
         private String outPinName = "";
         private String inPinName = "";
         private String serialName = "";
 
-        private String getI2cName(){
-            return i2cName ;
+        private String getI2cName() {
```
stefanhaustein: I wouldn't use getters for private variables, they seem to add boilerplate without much benefit here.

taartspi:
@FDelporte    I added the serial test. It has a work around as I opened an issue against ffm.  I would like us to merge this after comments. then I will pursue open issues against ffm

stefanhaustein (approved):
Sorry for the delay -- I somehow seem to have missed the review request. Don't hesitate to ping me if you wait for a response

pi4j-test/src/main/java/com/pi4j/test/Main.java
FDelporte: methods should have been public

pi4j-test/src/main/java/com/pi4j/test/Main.java
FDelporte: renamed

Issue #563, opened 2025-12-02
ffm PWM  channel other than 0 fail to export after reboot.


taartspi:
There are two statements that use chip rather than channel in PwmFFMHardware     

Reboot and repeated attempt to create a PWM on a channel > 0 will then work after two fails.  Happens as the first attempts each progress further through the initialization code.     

I will continue investigation of Smoke test fail along with these fixes.

taartspi:
In working on this as linuxfs sees less problems I looked at that code.  Well, now i remember when I completed linuxfs PWM for the Pi5 RP1, I added a delay when exporting so the new contexts and files could become persisted in storage prior to accessing them,  Something similar likely is needed in ffm

taartspi:
PR 564 fixes this

FDelporte:
Looking into the pull request, it's not clear if the delay is needed for the test implementation or required in all PWM use cases? Should the delay be part of the provider?

DigitalSmile:
I have updated the code with version of incremental watcher.

Though, I think it's better to use something like `WatcherService` from NIO2. I will create experimental PR

DigitalSmile:
@taartspi could you please double check the new implementation in this PR -> https://github.com/Pi4J/pi4j/pull/569 ?

taartspi:
Neither Frank nor I can repeat this error.  I will write it off to pilot error and close.

PR #564, opened 2025-12-08
PWM fix for initialization after reboot.   Updates to smoleTest main

taartspi:
Two problems after reboot.   The export needs some time to complete   I went through the same problem when i completed the linuxfs-pwm   Pi support forum Engineer agreed this was needed.  Other problem I missed a spot to handle the channel chip address changes

### DIFF
--- a/pi4j-test/src/main/java/com/pi4j/test/Main.java
+++ b/pi4j-test/src/main/java/com/pi4j/test/Main.java
@@ -31,11 +31,22 @@ import com.pi4j.io.gpio.digital.*;
 import com.pi4j.io.i2c.I2C;
 import com.pi4j.io.pwm.Pwm;
 import com.pi4j.io.pwm.PwmType;
+import com.pi4j.io.serial.*;
 import com.pi4j.io.spi.Spi;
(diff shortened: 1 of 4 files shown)


FDelporte:
Fixes #563

FDelporte (approved):

pi4j-test/src/main/java/com/pi4j/test/Main.java
```diff
@@ -138,21 +116,32 @@ public static void main(String[] args) throws Exception {
                 logger.info(helpString);
                 System.exit(31);
             } else {
-                logger.info("  !!! Invalid Parm " + args);
+                logger.info("  !!! Invalid Parm " + args.toString());
```
eitch, line 119: This toString() is useless, can just leave it out

eitch (approved):
Looks good

pi4j-test/src/main/java/com/pi4j/test/Main.java
taartspi: toString()  I was c;leaning up the many Intelli warnings. That was one of them

Issue #565, opened 2025-12-11
FFM Plugin: Changing the PWM dutyCycle with on() has no effect


taartspi:
Nothing happens when changing the duty cycle with:
```
pwm.on(dutyCycle);
```
But it does work with (but has unwanted on/off effect):
```
pwm.off();
pwm.on(dutyCycle);
```

FDelporte:
Fixed with PR #566

PR #566, opened 2025-12-11
#565: PWM on should be executed to handle changed values, even if already on

FDelporte:
Tested on Pi 4.


taartspi (approved):
Looks good. Thanks Frank

PR #569, opened 2025-12-13
#563 Added alternative implementation with WatcherService wrapper

### LINKED ISSUE #563: ffm PWM  channel other than 0 fail to export after reboot.

DigitalSmile:
Basic implementation of NIO2 wrapper for file creation waiting with fixed timeout.


plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/common/FileWatcher.java
```diff
@@ -0,0 +1,68 @@
...
+import java.nio.file.*;
+import java.time.Duration;
+import java.time.Instant;
+
+/**
+ * File watcher wrapper to wait for file is ready with defined timeout.
+ */
+public class FileWatcher implements AutoCloseable {
+    // in ms
+    private static final int TIMEOUT = 100;
```
eitch: Why not rename TIMEOUT_MS?

eitch: And possibly pass the timeout on construction?

eitch (approved):

taartspi:
code looks good.     I tried the PR that is merged, removes my direct delay and uses the existing waitForFile   This does not work.  The testcase creates the PWM and immediately call ON.  The On cannot access the file  open of the period,  fails access   Maybe the waitForFile File.access returns OK (0) before the file is fully created.  I will try this code next.

taartspi (commented):
code looks good. I tried the PR that is merged, removes my direct delay and uses the existing waitForFile This does not work. The testcase creates the PWM and immediately call ON. The On cannot access the file open of the period, fails access Maybe the waitForFile File.access returns OK (0) before the file is fully created. I will try this code next.

taartspi:
After reboot no pwm2 directory exists
  Fails first use in the smoketest   

[main] INFO com.pi4j.test.Main - Test PWM (freq: 1, dc: 50)
FAILED
Test failure: Error during call to method 'open' with data '[/sys/class/pwm/pwmchip0/pwm2/period, 1]': Permission denied (13)
The pwm2 directory was created  
Next run of the test worked correctly

PR #571, opened 2025-12-18
Refactor tests into separate classes

FDelporte:
I think this test project needed a full refactoring ;-)


taartspi (approved):
Looks good.

PR #572, opened 2025-12-19
ffm pwm does not call to the parent PWM to initialize values based on…

taartspi:
… the config object


eitch (approved):
That was missing, yes!

PR #573, opened 2025-12-19
Align FFM Provider names with other plugins

FDelporte:
As discussed during team call

### DIFF
--- a/pi4j-test/src/main/java/com/pi4j/test/type/PWMTestCase.java
+++ b/pi4j-test/src/main/java/com/pi4j/test/type/PWMTestCase.java
@@ -23,10 +23,16 @@ public class PWMTestCase extends TestCase {
             .newConfigBuilder(providerContext.getContext())
             .channel(PWM_CHANNEL)
             .pwmType(PwmType.HARDWARE)
+            .initial(dutyCycle)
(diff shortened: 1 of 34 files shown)


pi4j-test/src/main/java/com/pi4j/test/type/PWMTestCase.java
```diff
@@ -23,11 +23,17 @@ public static TestResult run(ProviderContext providerContext, int frequency, int
             .newConfigBuilder(providerContext.getContext())
             .channel(PWM_CHANNEL)
             .pwmType(PwmType.HARDWARE)
+            .initial(dutyCycle)
```
eitch, line 26: why are these added as part of the renaming?

eitch (approved):
looks otherwise good

pi4j-test/src/main/java/com/pi4j/test/type/PWMTestCase.java
FDelporte: my fault, uncommitted changes while testing yesterday, but anyway wanted :-)

taartspi (commented):
Thanks. And this one maybe your high water mark for number of files

PR #574, opened 2025-12-19
[WIP] Extending tests for clear steps and failed results 

FDelporte:
Trying to make the SmokeTests clear to understand and debug.


pi4j-test/src/main/java/com/pi4j/test/smokeTest/DigitalInputTestCase.java
```diff
@@ -0,0 +1,52 @@
+package com.pi4j.test.smokeTest;
```
eitch: no capital letters in package names =))

eitch (approved):
Except for the capital T in smokeTest, looks nice

pi4j-test/src/main/java/com/pi4j/test/smokeTest/DigitalInputTestCase.java
FDelporte: correct, fixed

taartspi (approved):
I take responsibility for the capital T.   Still old habits

FDelporte:
> I take responsibility for the capital T. Still old habits

@taartspi don't worry, the one @eitch was referring to, was one I introduced :-)

FDelporte:
Code compiles and builds on GitHub Actions succeed.
Will test first on Pi before merging, so keeping it in DRAFT / WIP for now.

taartspi:
changes i made  output pin uses a listener first. Then reads pin. 

PWM uses a listener .   The validation of monitor being low could fail since the PWM initial state is flashing   I repeated runs and saw once the count was 9, not ten once.    I did not look at the xtra test u added with higher values   

SPI  I need to see if Nicks FFM writeThenRead is complete and will test it      

Serial  When you moved this to the new class structure you removed a hack 'break' in the isOpen check   FFM did not implement the open() nor isOpen()  .I thought I documented that and opened an issue.  Could not find an issue (my bad)  opened a new one. 

@FDelporte

PR #575, opened 2025-12-21
Re-work the deferred delay class

stefanhaustein:
- Make it more suitable for immediate delays by making the set calls chainable and removing "deferred" from the name
- Add support for higher precision delays via parking and active waiting
- Improve conciseness by remove some redundancies between method names and the class name.

I hope it's fine to still do this kind of "API churn" as it was never releases (I'll delete the previous implementation when I have replaced all internal use cases)


taartspi (approved):
looks good      I don't remember what code uses the present class but a compile will flag them

PR #577, opened 2025-12-23
Shared writeThenRead interface

stefanhaustein:
Background / motivation: https://docs.google.com/document/d/1nlVF_9RwZm0OERCxPWvl-Qm6qypqfARxme8IQd7Dv6U/edit?usp=sharing

Open issues

- Any suggestions for a better name?
- In SPI, there is a 2nd delay parameter. Is that actually useful / needed for anything? 
- Are there any known use cases for SPI or I2C writeThenRead?

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/SerialCircuitIO.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/SerialCircuitIO.java
@@ -52,16 +52,16 @@ public interface SerialCircuitIO extends AutoCloseable {
     }
 
     /**
-     * This function writes bytes to the SPI device and then reads bytes from the device.
+     * This function writes bytes to the device and then reads bytes from the device.
      * Write data is taken from the 'writeBuffer' byte array from the given 'writeOffset' index to
      * the specified 'writeLength' (number of bytes).
      *
-     * <p>Data read back from the SPI device is then copied to the 'readBuffer' byte array starting
+     * <p>Data read back from the device is then copied to the 'readBuffer' byte array starting
      * at the given 'readOffset' using the 'readLength' (number of bytes).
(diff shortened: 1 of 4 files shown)


stefanhaustein:
Changed to only one delay, as discussed.

I have also changed the delay unit to nanoseconds as this seems to be relatively common in java apis (e.g. parking)...

pi4j-core/src/main/java/com/pi4j/io/SerialCircuitIO.java
```diff
@@ -0,0 +1,77 @@
...
+        writeThenRead(buffer, offset, length, 0, null, 0, 0);
+        return length;
+    }
+
+    default void writeThenRead(byte[] writeBuffer, byte[] readBuffer) {
+        writeThenRead(writeBuffer, 0, writeBuffer.length, 0, readBuffer, 0, readBuffer.length);
+    }
+
+    /**
+     * This function writes bytes to the SPI device and then reads bytes from the device.
```
eitch: Not SPI necessarily

pi4j-core/src/main/java/com/pi4j/io/SerialCircuitIO.java
```diff
@@ -0,0 +1,77 @@
...
+     * Write data is taken from the 'writeBuffer' byte array from the given 'writeOffset' index to
+     * the specified 'writeLength' (number of bytes).
+     *
+     * <p>Data read back from the SPI device is then copied to the 'readBuffer' byte array starting
```
eitch: dito

eitch (approved):
looks good otherwise

pi4j-core/src/main/java/com/pi4j/io/SerialCircuitIO.java
stefanhaustein: should be fixed now

PR #580, opened 2026-01-16
Remove serial implementations and related code

FDelporte:
As discussed in monthly call. Totally remove serial implementation from Pi4J.


eitch (approved):
looks good.

taartspi (commented):
All the java code looks good.  Appears you updated the smoketest doc. One thing, serial still referenced once in text last small chart.

PR #582, opened 2026-01-18
Implement setup permissions script

DigitalSmile:
Simple bash script to install permissions.
Tested in Ubuntu.


FDelporte:
This script is now available in the pi4j-os repository:
https://github.com/Pi4J/pi4j-os/blob/main/script/setup-permissions.sh

PR #585, opened 2026-01-21
#552 Added read delay to tx buffer, made adjustment to nano vs micro

### LINKED ISSUE #552: ffm.  SPI writeThenRead not implemented
This uses a single ioctl operation  presently implemented in linuxfs.

DigitalSmile:


DigitalSmile:
@taartspi please have a look at adjustments.
I have added readDelayNanos both to tx and rx buffer, adjusted nano seconds to micro seconds. That should work.

I don't think this `500` should be hardcoded in plugin side, you can adjust your method call with this values.

PR #589, opened 2026-01-21
Extra writeThenRead signature to avoid duplicate code

FDelporte:


pi4j-test/src/main/java/com/pi4j/test/smoketest/SpiTestCase.java
```diff
@@ -52,7 +52,7 @@ public static TestResult run(ProviderContext providerContext) {
     private static int readSpiRegister(Spi spi, int register) {
         byte[] data = new byte[]{(byte) (0b10000000 | register)};
         byte[] value = new byte[1];
-        spi.writeThenRead(data, value);
+        spi.writeThenRead(data, 0, value);
```
FDelporte, line 55: This will make it easier to test with different delays

eitch (approved):
looks good

PR #590, opened 2026-01-22
ffm i2cDirect line 104  out of bounds. smoketest remove second PWM te…

taartspi:
…st with freq too high. Enable spi test on FFM

### DIFF
--- a/pi4j-test/src/main/java/com/pi4j/test/Main.java
+++ b/pi4j-test/src/main/java/com/pi4j/test/Main.java
@@ -95,7 +95,6 @@ public class Main {
             I2CTestCase.run(providerContext),
             SpiTestCase.run(providerContext),
             PWMTestCase.run(providerContext, 1, 50, 10),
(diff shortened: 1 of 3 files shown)


taartspi:
smoketest changes   dropped the xtra PWM test.  Use the analyzer to see the PWM could pulse at 500 HZ, the input listener does not detect it.  If we want to further investigate the freq range I can  ???    Also enabled the SPI test on FFM as FFM now has the writeThenRead method.  This test on FFM  was failing when I pushed.   

I2C Direct in FFM.  This default was set as of today , fails smoketest  I2CDirect  line 104,  out of bounds.  I made a change and the testcase now succeeds.  Needs @DigitalSmile to look and decide if this is the correct fix.

eitch (approved):
For  me it makes sense, but Nick should comment too.

DigitalSmile:
This was just a typo, of course size should be > 0.

PR #591, opened 2026-01-23
move logger init into main. This allows setting the property to take …

taartspi:
…effect


taartspi:
Logger class only reads the properties one time during construction. So moved the LoggerFactory after setting the property

PR #598, opened 2026-01-27
Fix #594

### LINKED ISSUE #594: I2CRegisterDataReader     int readRegister(int var1);    byte -> conversion
Exception in thread "main" java.lang.IllegalStateException: WHO_AM_I register value -68 does not match expected value 188
at com.pi4j.drivers.sensor.environment.hts221.Hts221Driver.<init>(Hts221Driver.java:51)
Where the line above is int whoAmIValue = registerAccess.readRegister(15); (edited) 

So it looks like the signed byte -> conversion behavior changed for register reads (no longer matching java io read behaviour)?
I think this should be testable in smoke tests by setting the config or ctrl_meas register to a value > 127 and reading it back? (edited)

stefanhaustein:
Manually checked that the other implementations don't have this issue.

Tested manually via device demos. With some more effort, it should be possible to add a proper check in smoke tests by setting a bmp register to a value > 127 and reading it out.

### DIFF
--- a/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
+++ b/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
@@ -101,7 +101,7 @@ public class I2CDirect extends I2CBase<FFMI2CBus> {
 
     @Override
     public int readRegister(int register) {
-        return internalRead(new byte[]{(byte) register}, 1, new byte[0])[0];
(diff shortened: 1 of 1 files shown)


FDelporte (approved):

FDelporte:
@stefanhaustein regarding `add a proper check in smoke tests`, is this something you can add in `com.pi4j.test.smoketest.SpiTestCase.java` and/or `com.pi4j.test.smoketest.I2CTestCase.java` in the v4-testing branch?

DigitalSmile (approved):

DigitalSmile:
Good catch, I did not realize there could be unsigned values coming from driver side.
I guess we need to review other implementations of FFM to make sure we are doing right type conversion. Will do in a separate request.

PR #599, opened 2026-01-29
 smoketest update,  add debounce test to assist in debug

taartspi:
started adding a new TC to help with debug. 

Note.  I see that the linuxfs PWM test sometimes reports one additional flash.  Need to think on  this for different pwm.on() parameters to reduce the chances or this


taartspi:
@DigitalSmile Nick, the linuxfs reports the debounced event close to expected.  FFM with way late.

taartspi:
@FDelporte  IF you agree on this change please merge, if  u have comments I will update late tomorrow.

FDelporte (approved):

PR #600, opened 2026-01-30
debounce testcase has improved details of testcase outcome.  PWm test…

taartspi:
…, tweaked to close window where 10 or 11 flashes are counted. SPI test. Removed the test using seperate write and reds. Without the use of a dedicated GPIO to control CS from the application, write and reads in separate steps fail. Was the reason Stefan asked for writeThenRead implementation, it uses a single ioctl to perform both steps thereby keeping CS low. The testing of both means would be good to do with a chip that will support this


pi4j-test/src/main/java/com/pi4j/test/smoketest/SpiTestCase.java
```diff
@@ -37,20 +37,17 @@ public static TestResult run(ProviderContext providerContext) {
 
             // Read data from 0xD0 and check if the expected value is received
             byte idFromRead = readSpiRegisterUsingRead(spi, chipId);
```
FDelporte: idFromRead can be removed as it's not checked as a result?

pi4j-test/README.md
```diff
@@ -70,11 +70,12 @@ but your sensor may use different letter designations. Also, the are BMP sensors
...
-|:------------------|:-------------|:----------------:|:---:|:-----------|:------:|:---:|:------|
-| PWM | PWM0 |        12        | 18 | GPIO       | 16 | 23 | Green | 
-| Output | GPIO | 18 | 24 | GPIO | 22 | 25 | Blue  |
-| Input | GPIO | 36 | 16 | GPIO | 37 | 26 | White |
+| Test      | From RPi Pin | Number | BCM  | To RPi Pin | Number | BCM  | Color    |
+|:----------|:-------------|:------:|:----:|:-----------|:------:|:----:|:---------|
+| PWM       | PWM0         |   12   |  18  | GPIO       |   16   |  23  | Green    | 
+| Output    | GPIO         |   18   |  24  | GPIO       |   22   |  25  | Blue     |
+| Input     | GPIO         |   36   |  16  | GPIO       |   37   |  26  | White    |
+| Debounce  | GPIO         | 15     | 22   | GPIO       | 17     | 27   | Violet   |
```
FDelporte: I need to update the drawing :-)

pi4j-test/src/main/java/com/pi4j/test/smoketest/SpiTestCase.java
taartspi: done

taartspi:
@FDelporte    When u have time this is complete. Thanks for the artwork.   I had noted before the bmp i2c does work with a separate write then read transactions while the spi requires chip select low across separate write then read. The spec actually spells out the spi requirement. I2c test update is to test the writeThenRead uses restart between transfers. I know Nick updated to be two transfer in a single ioctl but it takes the logic analyzer to verify

taartspi:
@FDelporte Nick has two fixes for me to verify and our changes to this test will help.

Issue #604, opened 2026-02-02
In I2CDirect, in none of the 6 methods with an offset parameter, the offset parameter is referenced anywhere in the body


stefanhaustein:
The length parameter seems to be ignored a lot, too

stefanhaustein:
Might make sense to run a tool for unused parameters and similar issues over the whole ffm codebase?

FDelporte:
Hi, I'm trying to understand how this I2C implementation works. I reworked the internal methods in I2CDirect to include an offset parameter, so they are flagged as unused by SonarQube. This doesn't happen in the override methods...

Draft pull request is here https://github.com/Pi4J/pi4j/pull/605
This will need more work as it doesn't fix this problem yet, it only makes it more visible at this moment. 
@DigitalSmile can you assist me what more changes are needed?

<img width="691" height="245" alt="Image" src="https://github.com/user-attachments/assets/a0ab8d6a-7d34-47db-82b4-010cdfe6ddf5" />

stefanhaustein:
I re-ran the sensehat example and everything seems to work as expected now.

PR #605, opened 2026-02-03
Fix unused offset parameter in I2CDirect

FDelporte:
Related to #604


plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
```diff
@@ -11,126 +11,170 @@
...
 import java.nio.ByteBuffer;
 
 public class I2CDirect extends I2CBase<FFMI2CBus> {
-    private static final Logger logger = LoggerFactory.getLogger(I2CDirect.class);
-    private final IoctlNative IOCTL = new IoctlNative();
+    private final IoctlNative ioctl;
 
     public I2CDirect(I2CProvider provider, I2CConfig config, FFMI2CBus i2CBus) {
         super(provider, config, i2CBus);
+        ioctl = new IoctlNative();
```
stefanhaustein: Why this change? If a final variable can be assigned at declaration time, it probably should.

eitch (approved):
The code looks, but it's hard to know if its correct without fully testing it.

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
```diff
@@ -11,126 +11,170 @@
...
     public I2C initialize(Context context) throws InitializeException {
         return super.initialize(context);
     }
 
-    private byte[] internalRead(int size, byte[] buffer) {
+    private byte[] internalRead(byte[] buffer) {
+        return internalRead(0, buffer.length, buffer);
+    }
+
+    private byte[] internalRead(int offset, int length, byte[] buffer) {
```
stefanhaustein: Documenting what the parameter and returned value are supposed to do would probably be a good start here. Are they used for the same thing? Why have two byte arrays in read in the first place? 

The added offset parameter is now unused here....?

stefanhaustein (changes requested):

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
FDelporte: indeed not used, but now gets alerted by sonarqube. so easier to improve further but need feedback from @DigitalSmile as I'm not sure what code is doing here...

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
DigitalSmile: Agreed.

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
DigitalSmile: > indeed not used, but now gets alerted by sonarqube. so easier to improve further but need feedback from [@DigitalSmile](https://github.com/DigitalSmile) as I'm not sure what code is doing here...

I got the methods need to be override from other providers, I do not know which one good to use... That's more to interface declaration, rather than implementation.

DigitalSmile (changes requested):
Do we really need all of this methods in place?
For me it looks very confusing to have that many `offset` methods...

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
eitch: I don't agree =). Variables should be initialized in the constructor, this is standard Java.

stefanhaustein: reference? Regardless, isn't it generally, better to decouple stylistic changes from semantic changes so the review doesn't drown in noise?

stefanhaustein:
> Do we really need all of this methods in place? For me it looks very confusing to have that many `offset` methods...

A rule of thumb could be to look at the interfaces or abstract classes you are implementing and only implement the methods that don't already have a reasonable default there. 

Typically, if there are multiple methods with the same name, that's the method with the most parameters -- the others are typically just for user convenience

DigitalSmile:
> A rule of thumb could be to look at the interfaces or abstract classes you are implementing and only implement the methods that don't already have a reasonable default there.
>
> Typically, if there are multiple methods with the same name, that's the method with the most parameters -- the others are typically just for user convenience

It works if you are not owning those interfaces, but we do. Why we need all of them? Do they really used by users?

stefanhaustein:
> It works if you are not owning those interfaces, but we do. Why we need all of them? Do they really used by users?

I hope we can deprecate everything except for SerialCircuitIO after the next release, but that's really a different discussion. 

There should be a very small number of methods that don't have proper default implementations and these should be all that's needed here.

FDelporte:
@stefanhaustein back to the original issue of this pull request. did you report because you found broken functionality because of the unused parameters? if not, than we can indeed review all methods later to check if some can be deprecated.

stefanhaustein:
> @stefanhaustein back to the original issue of this pull request. did you report because you found broken functionality because of the unused parameters? if not, than we can indeed review all methods later to check if some can be deprecated.

Yes, I did run into driver issues exposed by the sense hat demo because of these bugs.

IMHO, there should be no question that ignoring an offset parameter -- leading to reading data from or writing data to the wrong position in an array absolutely has to be fixed no matter if the corresponding methods get deprecated at some point in the future or not?

taartspi:
Yes. Any method with offset parms need a proper implementation.  
I believe some methods have two offset parms one where the data written begins and the second   
where the read data is placed in the users buffer.  
Likely there are some pieces of the interface not being used. I think some of the signatures have a history from long ago and could be dropped.    @FDelporte could we depreciate signatures at a point release ?

taartspi:
@FDelporte   You asked about an example using buffers  
In the smoketest I2C.  This is not in a PR, but likely needed once this duscusion is finished.

//  i2c.writeThenRead(reg, wrData);
            // Read data from 0xD0 and check if the expected value is received
            byte[] register = new byte[]{0x00, 0x00,(byte)0xd0};
            byte[] readData = new byte[7];

            i2c.writeThenRead(register,2,1,0,readData,3,1);
            byte idFromReadThenWrite = readData[3];;

            //byte idFromReadThenWrite = wrData[0];

FDelporte:
@stefanhaustein can you test if these changes fix the problem with the sense hat you found?

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
```diff
@@ -11,126 +11,186 @@
...
+    private byte[] internalRead(byte[] buffer) {
+        return internalRead(0, buffer.length, buffer);
+    }
+
+    /**
+     * Read data from the I/O device into the provided byte array at the given offset and up to the specified data length (number of bytes).
+     * @param offset position in the buffer for the read data
+     * @param length max data length to be added to the buffer
+     * @param buffer the buffer to read the data
+     * @return byte array
```
stefanhaustein: How does the returned buffer differ from the "buffer" byte array handed in?

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
```diff
@@ -11,126 +11,186 @@
...
+     */
+    private byte[] internalRead(int offset, int length, byte[] buffer) {
+        var readBuffer = new byte[length];
```
stefanhaustein: This allocation seems to be unnecessary if the offset is 0?

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
```diff
@@ -11,126 +11,186 @@
...
+            return buffer;
         });
     }
 
     private int internalWrite(byte[] register, byte[] data) {
-        var buffer = new byte[register.length + data.length];
+        return internalWrite(register, 0, data.length, data);
+    }
+
+    private int internalWrite(byte[] register, int offset, int length, byte[] data) {
```
stefanhaustein: Probably would makes sense to document the parameters here. In particular, typically, offset and length come after the byte array, but here they seem to be for the 2nd array?

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
```diff
@@ -11,126 +11,186 @@
...
+    /**
+     * Write an array of byte values with given offset (starting position) and length in the provided data array.
+     * @param offset
+     * @param length
+     * @param data
+     * @return the number of bytes successfully written
+     */
+    private int internalWrite(int offset, int length, byte[] data) {
+        var writeBuffer = new byte[length];
+        System.arraycopy(data, offset, writeBuffer, 0, length);
```
stefanhaustein, line 108: The allocation seesm tobe unnecessary if offset = 0.

What is the motivation for the unusual parameter order?

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
```diff
@@ -11,126 +11,186 @@
...
+    /**
+     * {@inheritDoc}
+     */
     @Override
     public I2C initialize(Context context) throws InitializeException {
         return super.initialize(context);
     }
 
-    private byte[] internalRead(int size, byte[] buffer) {
+    private byte[] internalRead(byte[] buffer) {
```
stefanhaustein: This seems to be used only used once and doesn't seem to provide a significant readability improvement... Why not just delete this one?

stefanhaustein (approved):
Looks certainly better than the current state... 

Probably simplest to commit and take it from there? Will simplify testing on my side a lot O:)

FDelporte:
@DigitalSmile I know you prefered to solve the parameter use in the inherited methods, but can you still review this version please? It's the best approach I could come up with ;-)

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
DigitalSmile: > What is the motivation for the unusual parameter order?

If I understand the method signature in interfaces correct, when you are passing offset / length with `write` method you want to write not entire buffer. but rather ranged data `[offset; length)`.

stefanhaustein: I was asking why the parameter _order_ is different from other similar methods in the JRE or in Pi4j. Everywhere else, the order is buffer, offset, length (= buffer first). Here, it is offset, length, buffer (buffer last). This is confusing in particular when there are two buffers involved, creating ambiguity which buffer the refinement parameters are for.

I'd recommend to just adjust the parameter order for internal methods to be consistent with the public methods.

DigitalSmile: You comment was not referring to the method signature line

FDelporte: order of parameters has been changed

DigitalSmile (approved):

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
```diff
@@ -11,126 +11,200 @@
...
+     * {@inheritDoc}
+     */
     @Override
     public I2C initialize(Context context) throws InitializeException {
         return super.initialize(context);
     }
 
-    private byte[] internalRead(int size, byte[] buffer) {
+    private byte[] internalRead(byte[] buffer, int offset, int length) {
+        Objects.checkFromIndexSize(offset, buffer.length, length);
```
stefanhaustein, line 33: Now I get a crash instead of bad values. I think the two length parameters are mixed up here.

stefanhaustein: I'll address this in PR #608

Issue #597, opened 2026-01-27
DigitalInput Debounce


taartspi:
final DigitalInputConfig buttonConfig = DigitalInput.newConfigBuilder(pi4j)
       .id("BCM_" + PIN_BTN)
       .name("Button")
       .bcm(PIN_BTN)
       .pull(PullResistance.PULL_DOWN)
       .debounce(3000L)
       .build();  // build the config
final DigitalInput button = pi4j.create(buttonConfig);

//get notified whenever the button state changes
button.addListener(e -> {
    switch (e.state()) {
       case HIGH    -> { logger.info("Button was pressed!");
                        led.high();
                              }
       case LOW     -> { logger.info("Button was depressed!");
                        led.low();
                              }
       case UNKNOWN -> logger.info("Something unknown happened!!");
    }
});
When he releases/depresses the button, then the listener is triggered multiple times with HIGH. This is not good. Is the debounce not working?

Log:
https://pi4j.slack.com/files/U01KA0VEM4M/F0AAM1H17PW/untitled

Can this be of any use [@Nick Gritsenko](https://pi4j.slack.com/team/U08K3RURBLM)? with FFM and trace logging. I also see the problem with debounce not working as expected.
I added trace logging to this JBang example:
https://github.com/Pi4J/pi4j-jbang/blob/main/digital/Button.java#L40
See snippet above for longer log
[/dev/gpiochip0-event-detection-pin-22] TRACE com.pi4j.plugin.ffm.providers.gpio.FFMDigitalInput$EventWatcher - /dev/gpiochip0-event-detection-pin-22 - Detected new event: LineEvent{timestampNs=623536501068, id=2, offset=22, seqno=2, lineSeqno=2}
[/dev/gpiochip0-event-detection-pin-22] TRACE com.pi4j.plugin.ffm.providers.gpio.FFMDigitalInput$EventWatcher - /dev/gpiochip0-event-detection-pin-22 - Added to event list new state FALLING
[/dev/gpiochip0-event-detection-pin-22] TRACE com.pi4j.plugin.ffm.providers.gpio.FFMDigitalInput$EventWatcher - /dev/gpiochip0-event-detection-pin-22 - Total events: 1
Button changed in listener to: LOW
Button was pressed for the 1th time
[...]

PR #606, opened 2026-02-05
FFM digital input debounce improvements with extra smoke tests

FDelporte:
Related to #597

@taartspi can you please review if this logic should be correct for testing correct debounce behavior?
Unfortanetly, I get unreliable results, with different counts at each run...

Info 20260211 provided by Claude AI:

```
The DigitalInputDebounceCountTestCase (line 59) creates a digital input with a 100ms 
debounce (100_000L microseconds). The test verifies that rapid state changes within 
the debounce period are filtered out, expecting only 2 HIGH and 1 LOW events.

## Debounce Implementation Comparison

1. LinuxFS Plugin (LinuxFsDigitalInput.java)
No software debounce implementation
Relies entirely on the Linux kernel's sysfs GPIO interface (lines 110-116)
Simply monitors file changes via WatchService (lines 139-154)
The kernel itself may handle debounce at hardware level depending on the platform

2. FFM Plugin (FFMDigitalInput.java)
Attempts hardware debounce at lines 95-102
Sets GPIO_V2_LINE_ATTR_ID_DEBOUNCE attribute in the kernel (line 100)
Converts microseconds to nanoseconds: (int) debounce * 1000
No software fallback if hardware debounce isn't supported
Directly dispatches all events from kernel to listeners (lines 131-140)

3. GpioD Plugin (GpioDDigitalInput.java) ✅
Software debounce implementation at lines 129-143
Algorithm:
  Reads event from kernel
  Waits for debounce period while monitoring for new events
  If new events arrive within debounce window, updates to latest event and resets timer
  Only dispatches event after debounce period passes with no new events
Also filters duplicate states (lines 145-150)

4. PiGpio Plugin (PiGpioDigitalInput.java) ✅
Hardware debounce via pigpio library at lines 99-109
Uses gpioGlitchFilter function with steadyInterval
Capped at 300ms maximum (lines 102-106)

## The Problem

The FFM plugin passes the debounce value to the kernel GPIO subsystem but:
Not all GPIO chips support hardware debounce
There's no software fallback if hardware debounce fails
Events are dispatched immediately without any software filtering (line 138)

## Solution

The FFM plugin needs software debounce similar to the GpioD implementation. 
The logic should be added in the EventWatcher.run() method 
(starting at line 234 in FFMDigitalInput.java), between reading events from the kernel 
(lines 264-283) and processing them (line 286).

The implementation should:
Keep the hardware debounce configuration (lines 95-102) as it works on supported hardware
Add software debounce in the event processing loop (around line 286)
Track last event timestamp and filter out events within the debounce window
Only dispatch events after the debounce period has passed
```

### DIFF
--- a/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMDigitalInput.java
+++ b/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMDigitalInput.java
@@ -48,7 +48,7 @@ public class FFMDigitalInput extends DigitalInputBase implements DigitalInput {
     private final PollNative poll = new PollNative();
 
     private final String deviceName;
-    private final int pin;
+    private final int bcm;
     private final long debounce;
     private final PullResistance pull;
     private int chipFileDescriptor;
@@ -62,9 +62,9 @@ public class FFMDigitalInput extends DigitalInputBase implements DigitalInput {
 
     public FFMDigitalInput(String chipName, DigitalInputProvider provider, DigitalInputConfig config) {
         super(provider, config);
-        this.pin = config.bcm();
+        this.bcm = config.bcm();
         this.deviceName = "/dev/gpiochip" + config.bus();
-        this.debounce = (config.debounce() != null && config.debounce() >= 0) ? config.debounce() : DigitalInput.DEFAULT_DEBOUNCE;
+        this.debounce = (config.debounce() != null && config.debounce() >= 0) ? config.debounce() : 0;
         this.pull = config.pull();
         FFMPermissionHelper.checkDevicePermissions(deviceName, config);
     }
@@ -75,21 +75,22 @@ public class FFMDigitalInput extends DigitalInputBase implements DigitalInput {
         try {
             if (!canAccessDevice()) {
                 var posix = Files.readAttributes(Path.of(deviceName), PosixFileAttributes.class);
-                logger.error("Inaccessible device: '{} {} {} {}'", PosixFilePermissions.toString(posix.permissions()), posix.owner().getName(), posix.group().getName(), deviceName);
+                logger.error("Inaccessible device: '{} {} {} {}'",
+                    PosixFilePermissions.toString(posix.permissions()), posix.owner().getName(), posix.group().getName(), deviceName);
                 logger.error("Please, read the documentation <link> to setup right permissions.");
                 throw new InitializeException("Device '" + deviceName + "' cannot be accessed with current user.");
             }
-            logger.info("{}-{} - setting up DigitalInput Pin...", deviceName, pin);
-            logger.trace("{}-{} - opening device file.", deviceName, pin);
+            logger.info("{}-{} - setting up DigitalInput BCM...", deviceName, bcm);
+            logger.trace("{}-{} - opening device file.", deviceName, bcm);
             var fd = file.open(deviceName, FileFlag.O_RDONLY | FileFlag.O_CLOEXEC);
-            var lineInfo = new LineInfo(new byte[]{}, new byte[]{}, pin, 0, 0, new LineAttribute[]{});
-            logger.trace("{}-{} - getting line info.", deviceName, pin);
+            var lineInfo = new LineInfo(new byte[]{}, new byte[]{}, bcm, 0, 0, new LineAttribute[]{});
+            logger.trace("{}-{} - getting line info.", deviceName, bcm);
             lineInfo = ioctl.call(fd, Command.getGpioV2GetLineInfoIoctl(), lineInfo);
             if ((lineInfo.flags() & PinFlag.USED.getValue()) > 0) {
                 this.shutdownInternal(context());
-                throw new InitializeException("Pin " + pin + " is in use");
+                throw new InitializeException("BCM " + bcm + " is in use");
             }
-            logger.trace("{}-{} - DigitalInput Pin line info: {}", deviceName, pin, lineInfo);
+            logger.trace("{}-{} - DigitalInput BCM line info: {}", deviceName, bcm, lineInfo);
             var flags = PinFlag.INPUT.getValue() | PinFlag.EDGE_RISING.getValue() | PinFlag.EDGE_FALLING.getValue();
             var attributes = new ArrayList<LineConfigAttribute>();
             if (debounce > 0) {
@@ -98,7 +99,7 @@ public class FFMDigitalInput extends DigitalInputBase implements DigitalInput {
                     throw new InitializeException("Debounce value of " + debounce + " is too large");
                 }
                 var debounceAttribute = new LineAttribute(LineAttributeId.GPIO_V2_LINE_ATTR_ID_DEBOUNCE.getValue(), 0, 0, (int) debounce * 1000);
-                attributes.add(new LineConfigAttribute(debounceAttribute, 1L << pin));
+                attributes.add(new LineConfigAttribute(debounceAttribute, 1L << bcm));
             }
             flags |= switch (pull) {
                 case OFF -> 0;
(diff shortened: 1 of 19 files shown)


taartspi (approved):
Looks correct.    If u run the existing debounce test is it a success ?

DigitalSmile:
First of all, `Thread.sleep` with values less then 50ms does not make any sense, since JVM simply cannot sleep that low duration.

Change all sleeps to 100, and try again. You are getting different results, because it is calculated in different thread.

FDelporte:
Thanks for the feedback @DigitalSmile, you're right, as always ;-) 

I now get consistent result, not sure if test really fails or my assumptions regarding expected behavior are wrong:

```
[main] INFO com.pi4j.test.Main - Test Digital Debounce
FAILED
Incorrect state change counts detected, expected high and low 2/1, but got 3/2
```

DigitalSmile:
> I now get consistent result, not sure if test really fails or my assumptions regarding expected behavior are wrong:

Don't really get your idea... You are calling `gpioOutTest.high();` in your code 3 times and `gpioOutTest.low();` 2 times, so it should be `3/2`

FDelporte:
> high 3 times

I'm trying to switch a few times from high to low within the debounce time to test if they get detected. Maybe I'm trying something which is not possible from code...

DigitalSmile:
> a few times from high to low within the debounce time

I think you cannot test debounce like this or the code you wrote already covers it... You are calling high/low several times if the wire contact between pins is not good it should be covered by debounce provided without any other adjustments.

But I am not expert in RPI hardware by any means :) May be Tom can add more details.

taartspi:
@FDelporte my first debounce test leniently makes sure it works.  I think this additional test is a good idea.  I will have some time tomorrow, is there any work on this test i can help with ?

FDelporte:
@taartspi would be nice if you could review "the flow". not sure I truly made the correct assumptions regarding the debounce values versus sleep.

taartspi:
I agree with your assumptions    Running with -p linuxfs the gpiod provider is successful.  I will spend more time reading the ffm logged data.   .

taartspi:
@FDelporte Morning  I flipped this PR to draft in case we throw away what i pushed.  This is to use the logic analyzer to measure debounce   The listener drives a gpio high, the analyzer capture the GPIO we drove high to start the test, and the added pin driven high when the listener runs.  I was lenient in the simple debounce test as the wait varies,  I see variation doing it this way as well.  I plan to also run this against the GPIOD provider to see it's variance.    Your logic looked correct,  I can run yours more tomorrow and I will have time to experiment more with this testcase tomorrow.  comment if u look at what i pushed.   Tom..

taartspi:
@nick    Nick, I would like more of your help to understand the testcase counting DigitalInputDebounceCountTestCase and DigitalInputDebounceTimeTestCase working on linuxfs and count failing ffm and the Time test fails some of the time.   Are we testing outside the limits of FFM, or is there a problem in the test or FFM ?  thanks

DigitalSmile:
> working on linuxfs and count failing ffm

Tom, why do you think it is failing? I explained to Frank here https://github.com/Pi4J/pi4j/pull/606#issuecomment-3854269643 that it should be 2/3 since this is how many times high/lows are called in the code. The debounce time of 100_000 microseconds cannot be catched by Thread sleeping since the minimum time to sleep is not guaranteed by JVM (of course if it is not running on RTOS). So writing `Thread.sleep(any_value)` or `LockSupport.parkNanos(any_value)` does not sleep exact `any_value` it could be much more depending the OS and hardware.

That is why I wrote previously it does not make much sense to test debounce between two pins...

Again, I am not an expert, but I thought debounce is used to prevent bad connection to trigger false events?

FDelporte:
@DigitalSmile some of the high/lows are called within the debounce time. With other plugin we get 1/2 as a result. Also testing with jbang and physical presses on a button shows it doesn't work as expected. Maybe we need extra debounce implementation in software if kernel doesn't support this correctly?

DigitalSmile:
> Maybe we need extra debounce implementation in software if kernel doesn't support this correctly?

That's what I wrote in the related issue. But we are wandering in shadows without understanding how it is supposed to work.

DigitalSmile:
> some of the high/lows are called within the debounce time.

Btw, that's not true. Debounce time is `100_000` microseconds, while sleep time between calls is _at least_ `50` milliseconds

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
```diff
@@ -101,7 +101,7 @@ public int write(byte[] data, int offset, int length) {
 
     @Override
     public int readRegister(int register) {
-        return Byte.toUnsignedInt(internalRead(new byte[]{(byte) register}, 1, new byte[0])[0]);
+        return Byte.toUnsignedInt(internalRead(new byte[]{(byte) register}, 1, new byte[1])[0]);
```
eitch, line 104: Is it a good idea to put this in this PR? Perhaps it should be in its own PR and we need to merge it soon?

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMDigitalInput.java
```diff
@@ -64,7 +64,7 @@ public FFMDigitalInput(String chipName, DigitalInputProvider provider, DigitalIn
         super(provider, config);
         this.pin = config.bcm();
         this.deviceName = "/dev/gpiochip" + config.bus();
-        this.debounce = config.debounce();
+        this.debounce = (config.debounce() != null && config.debounce() >= 0) ? config.debounce() : DigitalInput.DEFAULT_DEBOUNCE;
```
DigitalSmile: We should not pass default debounce, there are drivers which does not support it officially, and exception will occur while configuring pin. So if no debounce present - leave it as 0

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMDigitalInput.java
```diff
@@ -278,13 +290,50 @@ public void run() {
                             if ((event.id() & this.pinEvent.getValue()) != 0) {
                                 var pinEvent = PinEvent.getByValue(event.id());
                                 logger.trace("{} - Added to event list new state {}", Thread.currentThread().getName(), pinEvent);
```
DigitalSmile: There should be different log message...

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMDigitalInput.java
```diff
@@ -246,9 +251,16 @@ public void run() {
                     if (pollData == null) {
                         var duration = timestamp.until(Instant.now()).toMillis();
                         logger.trace("{} - No events detected: polling timeout at {} (took {}ms)", Thread.currentThread().getName(), timestamp, duration);
-                        // timeout happened, process all left events, update timestamp
-                        eventProcessor.process(eventList);
-                        eventList.clear();
+                        // timeout happened, process pending debounced event if any
+                        if (lastDebouncedEvent != null && debounceNs > 0) {
```
DigitalSmile: Why do we need this if block? The event list should be empty if timeout occurred?

DigitalSmile (changes requested):
I guess we also need to add one unit test with debounce timing, equivalent of https://github.com/Pi4J/pi4j/blob/develop/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/unit/GPIOTest.java#L139

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMDigitalInput.java
FDelporte: Tried to remove it, but then the test fails again... It's needed for:

```
The problem is that with debouncing enabled, the last event sits in lastDebouncedEvent waiting for the debounce period to pass, but if no new events arrive, it never gets dispatched.
```

But needed a fix to use kernel timestamps consistently.

FDelporte:
> I guess we also need to add one unit test with debounce timing, equivalent of https://github.com/Pi4J/pi4j/blob/develop/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/unit/GPIOTest.java#L139

@DigitalSmile extra test got added

DigitalSmile (approved):

eitch (approved):

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CDirect.java
FDelporte: I noticed this while fixing the unused parameters, see https://github.com/Pi4J/pi4j/pull/605 and decided to included it also in this merge request while trying out the smoke tests... indeed, not the ideal approach ;-)

eitch: Well it got merged soon, so all's good =))

PR #607, opened 2026-02-12
smoketest wiring. If the sensor supports both i2c and spi, the CS lin…

taartspi:
…e controls which interface is active


FDelporte:
@taartspi are you sure about this? for my sensor to work on I2C, I had to remove the CS connection...
Can you add the link to the datasheet?

FDelporte (approved):

PR #608, opened 2026-02-12
Fix index/size check issue

stefanhaustein:


stefanhaustein:
Related to issue #604

PR #632, opened 2026-04-30
update smoketest readme tables to include LED# as defined in EasyEDA …

taartspi:
…design


FDelporte (approved):
Checked with current status of PCB design

Issue #613, opened 2026-02-18
Re-design PWM API for more accuracy


mstahv:
I spent quite a bit of time figuring out how to use PWM APIs to control a small "standard" micro servo motor. Finally I figured out that the current AP probably works for dimming leds, it is just incompatible with this kind of signals. Accuracy of percantage given as integer just isn't enough. I made the pi4j's PWM API (snapshot build of 4.0.0, linuxfs), somewhat work, but I only was able to adjust it to couple of specific locations.

Here is the PWM "specification" for my servo motor:

 * pulse every 20ms (~ 50Hz)
 * Pulse width of ~ 0.5ms = 0 degrees
 * Pulse width of ~ 2.4ms = 180 degrees

Now with a bit of math I can only set the servo to only about 10 different positions with the current pi4j API.

I was looking for a while to adjust the API, and noticed that the internal LinuxPwm would actually have some tooling, but making patch started to feel too complex. I'd maybe even suggest to rewamp those altogether: have some sort of lower level API with nanoseconds (like the HW via sysfs is adjusted) and then some additional helper classes for frequancy and proportional settings of different kinds (of the whole period and of specified min/max values, which would fit for e.g. servos). There could still be a common interface for both software and hardware pwm

Finally my solution is quite simple (and it accuratly admusts an old chain saw carborator 😎), with a helper class that just writes to system fs. Maybe this will help if somebody gets a good vision on how to implement this in pi4j (this now is only for single channel and only hardware pwm).

This is what I wrote to config.txt:

```
dtoverlay=pwm,pin=12,func=4
```

PwmChip class (couple of helpers here that should get into default methods of an interface):

```
public class PwmChip {
    
    private final int chipNumber;
    private final int channel;
    private final String basePath;
    private boolean exported = false;
    
    /**
     * Creates a new PWM chip controller for the specified chip and channel.
     * 
     * @param chipNumber The PWM chip number (typically 0 for /sys/class/pwm/pwmchip0)
     * @param channel The PWM channel (typically 0 for pwm0)
     */
    public PwmChip(int chipNumber, int channel) {
        this.chipNumber = chipNumber;
        this.channel = channel;
        this.basePath = "/sys/class/pwm/pwmchip" + chipNumber;
    }
    
    /**
     * Exports the PWM channel, making it available for configuration.
     * Equivalent to: echo 0 > export
     * 
     * @throws IOException if the export operation fails
     */
    public void export() throws IOException {
        if (exported) {
            return;
[...]

PR #646, opened 2026-06-09
Change frequency to a double (nullable Double in config)

stefanhaustein:
To be consistent with duty cycle and to enable 1.5 Hz or similar

EDIT: See Issue #613 for background / discussion


eitch (approved):

stefanhaustein:
Thanks for the quick review @eitch 

You have added two more reviewers; should I wait for approvals from both or one of them or is this intended as an FYI to make them aware of the change?

eitch:
@stefanhaustein a little of both. But Frank already said he doesn't have time. For me this is otherwise fine, so let's merge it. If @taartspi has an issue, we can revisit it.

Issue #649, opened 2026-06-11
Cleanup: What's obsolete now?


stefanhaustein:
I think one important question here is: What is actually intended to be user-visible in Context?

There are some properties that seem to mostly serve internal purposes... Iterating all of them:

- config: No documentation at all. Looks like a generic map. I have never used this. Is this something that is used by users and needs to be preserved?
- properties: No documentation at all. Looks like a generic map. I have never used this. Is this something that is used by users and needs to be preserved?
- providers: I understand what this is and we need to preserve this if we want to preserve providers, which is under discussion elsewhere.
- registry: Manages open devices and prevents re-using buses and pins. We will need to preserve the concept, but does it need to be user visible? Or is it user visible only for internal purposes?
- platforms: I have no idea what this is and there is no documentation at all. It looks like yet another generic map to me. I have never used this. Is this something that is used by users and needs to be preserved?

Not sure how to best figure out what we could remove from the user visible API?

Is for any of this clear that it had served a purpose before FFM-only and is now obsolete? 

How do we figure this out?
Does anybody remember anything about these properties @eitch @FDelporte ?

Would it be safe to assume that if something is not used in examples, it's probably not intended for external use?

eitch:
I am certain we can remove platforms as this was an earlier idea to support other platforms, i.e. Raspberry Pi and others, but that is not necessary thanks to FFM.
The properties and config i don't know. Probably not necessary anymore, unless there is a config that can be set and is used elsewhere.

taartspi:
'Would it be safe to assume that if something is not used in examples, it's probably not intended for external use?'  

I think not a safe assumption. Although there are many entries under https://www.pi4j.com/examples/ their intentions were not component testing, only to demonstrate something. The individual examples throughout the pi4j web pages use only enough features to explain/demonstrate

stefanhaustein:
Rough plan:

1. Remove Platforms
2. See if we can remove properties and config by trying to remove them (they seem to be too tangled to remove separately)
3. Move Runtime / Registry functionality directly into DefaultContext, as suggested by @eitch elsewhere earlier
4. See if we can get the providers discussion resolved in the meantime

Does that make sense?

eitch:
I like this plan

stefanhaustein:
[...]

PR #650, opened 2026-06-12
Remove Platforms

stefanhaustein:
As suggested in the cleanup thread: https://github.com/Pi4J/pi4j/issues/649

This also opens up the DefaultContext for inheritance to simplify experimentation with provider-free Context implementations.


stefanhaustein:
Any concerns? It would be great if we could keep the ball rolling and get some momentum here...
I'll try to remember to use a dedicated fork for the next cleanup, so I don't block my main fork for this stuff... O:)

eitch:
I haven't had time to look at it... Sorry, family time, jogging, etc. Will look at it on Monday.

PR #651, opened 2026-06-13
Test framework

### LINKED ISSUE #643: Automation software test framework
Goals:

1) Provide solid unit testing framework for the simple method testing with expected vs actual results using mocks.
2) Provide low level interop testing of FFM models (alignment, types, native->java and vice versa conversion routine).
3) Provide integration tests using mock linux echo drivers for that.
4) Provide regular performance testing using integration drivers and JMH.

Prerequisites:

1) Should work in CI and pass all the tests automatically.
2) Should work in local developer environment in Linux and MacOS/Windows (with containers).
3) Should support amd64 and arm64 architectures.
4) Should publish results of constant performance testing (e.g. run once a week on develop branch if there are any changes and per PR before merge)

@eitch @stefanhaustein do not hesitate to add more details :)

DigitalSmile:
# Pi4J testing framework for the win

## Overview

Pi4J lacks proper tests within it's usage. Not only unit in core tests (which are good to test some basic logic), but provider unit, integration and performance tests. Moreover, new FFM API relies on the kernel, that is run within Linux OS. That's essentially means we have to keep an eye on what is changing in kernel driver/syscall development. To keep all this in place, I propose the following framework.

TLDR:
1) Please, write unit/model/integration/performance tests alongside with your changes to code.
2) Unit tests are run at every build - they provide mocks for every hardware iteration.
3) Model tests are run at every build and intended to check if C structs are mapped to Java objects correctly.
4) Integration tests are run in Linux only environment at each build (CI included). I wrote 4 mock kernel drivers to mimic the actual hardware: GPIO, I2C, PWM and SPI. They are simple drivers that echoes what is written. This branch build is already enabled.
5) Performance tests are run bi-weekly every 1 and 15 day of month and put results into `jmh-results` folder. There is an index.html page there to view and compare results (probably we will move it to pi4j website).

## Unit tests

~~We need to use mock approach to test raw methods, that they works as expected (e.g. Mockito in FFM https://github.com/Pi4J/pi4j/tree/develop/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/mocks).
Several layers of unit testing approach:~~
- ~~core unit tests, located at https://github.com/Pi4J/pi4j/tree/develop/pi4j-core/src/test/java/com/pi4j~~
- ~~FFM and any other provider (e.g. https://github.com/Pi4J/pi4j/tree/develop/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/unit)~~

~~In this scope, we need to remove `pi4j-test` module and move corresponding unit  and integration tests to core.~~

It's better to leave unit tests here for now.

## Model tests

Model tests are needed to make sure any interop structure are not broken.
- FFM model tests (https://github.com/Pi4J/pi4j/tree/develop/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/model)

Same should be applied to any other plugins we offer.

### Todo

- [ ] ~~Add model tests to pi4j-core (if necessary)~~

## Integration tests

This tests should check if interop and syscalls are not broken. Main idea is to write simple mock (or echo) drivers for each provider and test syscalls. I have created some drivers FFM provider within this PR:

- SPI interface
    - driver - https://github.com/Pi4J/pi4j/tree/test-framework/plugins/pi4j-plugin-ffm/src/test/native/spi
    - tests - https://github.com/Pi4J/pi4j/blob/test-framework/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/SPITest.java
- I2C
    - driver - https://github.com/Pi4J/pi4j/tree/test-framework/plugins/pi4j-plugin-ffm/src/test/native/i2c
    - I2C with ioctl - https://github.com/Pi4J/pi4j/blob/test-framework/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/I2CDirectTest.java
    - I2C with file - https://github.com/Pi4J/pi4j/blob/test-framework/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/I2CFileTest.java
    - I2C with SMBus - https://github.com/Pi4J/pi4j/blob/test-framework/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/I2CSMBusTest.java
- PWM
    - driver - https://github.com/Pi4J/pi4j/tree/test-framework/plugins/pi4j-plugin-ffm/src/test/native/pwm
    - tests - https://github.com/Pi4J/pi4j/blob/test-framework/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/PwmTest.java
- GPIO
    - driver - https://github.com/Pi4J/pi4j/blob/test-framework/plugins/pi4j-plugin-ffm/src/test/native/gpio/gpio-mock.c
    - tests - https://github.com/Pi4J/pi4j/blob/test-framework/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/GPIOTest.java
- ~~Serial **TBD**~~

This drivers requires a simple build with `make` and current loaded kernel headers (I have provided comments with source files, though there are really very simple). We can think of expanding this tests to all recent stable linux kernels in future.
Each driver is built and loaded with the test in java and removed upon test completion. This works in linux machines only, so a special annotation is used not to run them on other systems.

### Todo

- [x] Create SPI driver and fix tests
- [x] Create PWM driver and fix tests
- [x] Create I2C driver and fix tests
- [x] Create GPIO driver and fix tests
- [ ] ~~Create Serial driver and fix tests~~

## Performance tests

Since we are planning to expand the support from RaspberryPi only to every linux driven SBC, It is very possible, that pi4j will be used in high concurrent and/or highload environment. So we have to control and measure periodically how our changes affects the performance of interop and overall software responsiveness. I suggest to use the same approach as in integration testing - use simple echo/mock drivers to measure how fast the communication go. JMH tests for FFM provider are located here -> https://github.com/Pi4J/pi4j/tree/test-framework/plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/jmh
They include:

- GPIO performance tests
- PWM performance tests
- I2C performance tests
- SPI performance tests
- ~~Serial performance tests~~

We have to think on comparing ourselves to other / similar libraries.

### Todo

- [x] Create FFM API tests for GPIO (regression and performance degradation)
- [x] Create FFM API tests for PWM (regression and performance degradation)
- [x] Create FFM API tests for SPI (regression and performance degradation)
- [x] Create FFM API tests for I2C (regression and performance degradation)
- [ ] ~~Create FFM API tests for Serial (regression and performance degradation)~~
- [x] Create comparing performance tests
- [x] Run tests periodically and publish results


.github/workflows/ci-pull-request.yml
```diff
@@ -17,11 +17,10 @@ jobs:
           sudo apt-get remove --purge man-db
           sudo apt-get install --yes build-essential linux-headers-$(uname -r) sudo libi2c-dev linux-modules-extra-$(uname -r)
 
-      - name: Set up Zulu JDK 25
-        uses: actions/setup-java@v4
+      - name: Install JDK25
```
eitch: Why switch to SDKMan? I think using actions/setup-java is perfectly fine, also since your selected action has nod had any updates in ages.

.github/workflows/ci-pull-request.yml
```diff
@@ -47,7 +46,7 @@ jobs:
         env:
           MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
           MAVEN_PASSWORD: ${{ secrets.OSSRH_TOKEN }}
-        run: mvn clean package
+        run: sudo -u $USER bash -c 'source ~/.sdkman/bin/sdkman-init.sh && ./mvnw clean package'
```
eitch: why do we need sudo? I think we should make sure we can stick to mvn clean package, no?

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CFile.java
```diff
@@ -109,7 +109,7 @@ public int readRegister(int register, byte[] buffer, int offset, int length) {
      */
     @Override
     public int writeRegister(int register, byte b) {
-        return i2CBus.execute(this, (i2cFileDescriptor) -> FILE.write(i2cFileDescriptor, new byte[]{(byte) register, b}));
+        return i2CBus.execute(this, (i2cFileDescriptor) -> FILE.write(i2cFileDescriptor, new byte[]{(byte) register, b})) - 1;
```
eitch, line 112: What's with this?

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CFile.java
```diff
@@ -120,7 +120,7 @@ public int writeRegister(int register, byte[] data, int offset, int length) {
         var buffer = new byte[data.length + 1];
         buffer[0] = (byte) register;
         System.arraycopy(data, 0, buffer, 1, data.length);
-        return i2CBus.execute(this, (i2cFileDescriptor) -> FILE.write(i2cFileDescriptor, buffer));
+        return i2CBus.execute(this, (i2cFileDescriptor) -> FILE.write(i2cFileDescriptor, buffer)) - 1;
```
eitch, line 123: What's with this?

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/spi/FFMSpi.java
```diff
@@ -102,8 +102,10 @@ public int transfer(byte[] write, int writeOffset, byte[] read, int readOffset,
         if (read != null) {
             System.arraycopy(readBytes, 0, read, readOffset, numberOfBytes);
             logger.trace("{} - Read buffer: {}", path, HexFormatter.format(read));
+            return readBytes.length;
```
eitch: the null checks on the read and write buffer are useless as we reference them before anyhow already. I know not part of the fix, but clean ups are always welcome

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/spi/FFMSpi.java
```diff
@@ -102,8 +102,10 @@ public int transfer(byte[] write, int writeOffset, byte[] read, int readOffset,
...
+        } else {
+            return spiTransfer.getTxBuffer().length;
```
eitch: The JavaDoc doesn't reflect what this method returns.

plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/I2CSMBusTest.java
```diff
@@ -112,7 +114,7 @@ public void testI2CWriteReadRegister() throws InterruptedException {
     }
 
 
-    @Test
+    //@Test
```
eitch: i don't like commented out tests like these

FDelporte (approved):

plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/I2CSMBusTest.java
```diff
@@ -22,54 +21,57 @@
...
-            var setupScript = new ProcessBuilder("/bin/bash", "-c", "sudo " + scriptPath.toFile().getAbsolutePath()).start();
-            var result = setupScript.waitFor();
-            if (result != 0) {
-                var username = System.getProperty("user.name");
-                var errorOutput = new String(setupScript.getErrorStream().readAllBytes());
-                fail("Failed to cleanup I2C Test: \n" + errorOutput + "\n" +
-                    "Probably you need to add the I2C bash script to sudoers file " +
-                    "with visudo: '" + username + " ALL=(ALL) NOPASSWD: " + scriptPath.toFile().getParentFile().getAbsolutePath() + "/'");
-            }
+        var scriptPath = Paths.get("src/test/resources").toFile().getAbsoluteFile();
```
eitch: This code seems to be duplicated a few times...

plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/jmh/GPIOInputPerformanceTest.java
```diff
@@ -0,0 +1,85 @@
...
+@BenchmarkMode({Mode.AverageTime})
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+public class GPIOInputPerformanceTest {
+
+    private Context pi4j;
+    private DigitalInput pin;
+
+    @Setup(Level.Trial)
+    public void setup() throws InterruptedException, IOException {
+        var scriptPath = Paths.get("src/test/resources").toFile().getAbsoluteFile();
```
eitch: again a lot of duplication

eitch (commented):
I love the idea of the kernel mocks for the different IOs, this makes testing everything really easy on a desktop. I would just prefer that the tests are not run as sudo. All the setup should be done before, and then run the normal tests. Furthermore i see code changes which are not part of the test framework itself and are for me hidden changes, without a discussion. They might be correct, i just can't tell.

.github/workflows/ci-pull-request.yml
DigitalSmile: The answer is tight with SDKMAN. As you can see in jobs run before we are creating groups to work with IO. To handle it right in current session we need to do two things:
1) Update session with `source ~/.sdkman/bin/sdkman-init.sh` makes java available for the user immidiately. I did not find any solution for setup-java :(
2) `sudo -u $USER bash -c` does not run anything under root, it's running the command under the github runner. With all updated groups and this is the only way to get them work with current job run.

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CFile.java
DigitalSmile: Same as above. Fix for tests to be green.

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CFile.java
DigitalSmile: This fix was made in order to tests runs green. the return value should be number of byte written. But our implementation counts register as a part of buffer. So if you are writing to register `1` value `1` it reports 2 bytes written, instead of 1.

plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/I2CSMBusTest.java
DigitalSmile: My bad, will fix.

DigitalSmile:
> All the setup should be done before, and then run the normal tests.

Well, I put all setup as a part of testing just not to mess with XML and maven. We can move it somehow, but I do not think it is better then calling it from the code.
Root is needed for kernel operations `insmod` and `rmmod`. They must be run by root. That is why I made scripts and a short message how to make it work.

Next iteration of testing framework will be `test-containers` and probably we can get rid of this. But for now this is the only way.

.github/workflows/ci-pull-request.yml
eitch: About point 2: You are right didn't look properly.
About point 1: I would expect that installing java with actions/setup-java does just that:
```
steps:
- uses: actions/checkout@v6
- uses: actions/setup-java@v5
  with:
    distribution: 'zulu' # See 'Supported distributions' for available options
    java-version: '25'
- run: java HelloWorldApp.java
```
directly from the docs. What here doesn't work then in your use-case? And did you try with v5? I'm not sure i like using such an old action

DigitalSmile: Will try new version.

DigitalSmile: I checked once again and it did not work. There is an issue with workaround, which is not better than using sdkman. -> https://github.com/actions/setup-java/issues/997

```
**NOTE:** On Ubuntu runners, commands executed via `sudo` do not inherit the `JAVA_HOME` and `PATH` set by `setup-java` and will fall back to the runner image's system-default JDK.
```

Why not to use sdkman?

plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/jmh/GPIOInputPerformanceTest.java
DigitalSmile: Done

plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/I2CSMBusTest.java
DigitalSmile: Done

plugins/pi4j-plugin-ffm/src/test/java/com/pi4j/plugin/ffm/integration/I2CSMBusTest.java
DigitalSmile: Done

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/spi/FFMSpi.java
DigitalSmile: There are a lot of javadoc outdated... I put some fixes, but let's do the clean up in a separate PR.

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/spi/FFMSpi.java
DigitalSmile: Done

eitch (approved):
Thanks for all the changes, it looks good!

PR #663, opened 2026-06-19
#654 Added an initial state to FFMDigitalOutput

### LINKED ISSUE #654: FFM         DigitalOutput         .initial(DigitalState.HIGH);   GPIO pin pulled low, then goes high
If the pin's use is a chip reset,   and reset is enabled with a low:
     Upon restarting an application the reset lines go low and then high during DigitalOutput creation.   Any past chip register state is lost based upon the chips POR operation.

@IAmNickNack    Nick, I don't know if the previous providers had the same behaviour but it presents a problem.

DigitalSmile:


eitch (approved):

taartspi:
Good addition @DigitalSmile ,  works great.

PR #666, opened 2026-06-19
#654 Added a chunked size SPI transfers

### LINKED ISSUE #475: [linuxfs] How to send large amounts of data over SPI
In the past when using pigpio - a fix was applied to break up large data into chunks: https://github.com/Pi4J/pi4j/issues/383

I am using linuxfs and seeing the following error:
2025-08-11 11:08:12,964 ERROR [LinuxFsSpi:332] () Could not write SPI message. ret -1, error: 90

I see when using linuxfs buffer size can be configured: https://www.pi4j.com/documentation/providers/linuxfs/

What is the proper fix going forward ?

1.  Should the buffer size simply be increased to the amount needed
2. Or should there be a similar paging mechanism to support large data using fixed buffer size ?

DigitalSmile:
@taartspi please have a look and test on a real hardware


eitch (approved):

taartspi (commented):
@DigitalSmile Tested the write and writeThenRead with bufsize less then the number bytes I was sending, worked correct;y.

PR #673, opened 2026-06-19
Support 16GB Compute Module 5 variant

pjbosco:
Updated BoardModel.COMPUTE_5's 'boardCodes' to reflect reality.

The CM5 and CM5 Lite exist today in 2GB, 4GB, 8GB, and 16GB variants with revision code prefix b, c, d, and e, respectively.

Removed the revision codes prefixed with 'a' (1GB) and added revision codes prefixed with 'e' (16GB) to allow for automatic BoardModel detection of CM5/CM5 Lite 16GB variants.

Updated BoardModel.COMPUTE_5's 'remarks' to include a link to official documentation of the current known revision codes.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/boardinfo/definition/BoardModel.java
+++ b/pi4j-core/src/main/java/com/pi4j/boardinfo/definition/BoardModel.java
@@ -244,7 +244,7 @@ public enum BoardModel {
         List.of(1024 * 1024, 2048 * 1024, 4096 * 1024, 8192 * 1024)),
     // https://www-cnx--software-com.cdn.ampproject.org/c/s/www.cnx-software.com/2024/11/27/raspberry-pi-cm5-broadcom-bcm2712-soc-16gb-lpddr4-ecc-memory/?amp=1
     COMPUTE_5("Compute Module 5", STACK_ON_COMPUTER,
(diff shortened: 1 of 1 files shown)


DigitalSmile:
@pjbosco Hello! Thanks for the PR! Are the deleted items still used with already manufactured and sold boards? Probably with older firmware? My concern is if we can break users using old codes.

pjbosco:
> @pjbosco Hello! Thanks for the PR! Are the deleted items still used with already manufactured and sold boards? Probably with older firmware? My concern is if we can break users using old codes.

That's a good question. I have not been able to find any mention of 1GB CM5 or CM5 Lite boards and Raspberry Pi's own - albeit admittedly incomplete - documentation does not list a 1GB CM5 or CM5 Lite (would use the 'a' prefix). If we wanted to be completely safe, we could add the 'a' prefixed revision codes back in, but I doubt it had ever been used in the first place. 

The standard Raspberry Pi 5 has a 1GB RAM variant, but that is a different BoardModel anyway.

DigitalSmile:
>  If we wanted to be completely safe, we could add the 'a' prefixed revision codes back in, but I doubt it had ever been used in the first place.

Makes sense. Let's just go with this fixes. Thank you!

PR #676, opened 2026-06-24
Smoketest   Changes to core context required testcase changes. Deleti…

taartspi:
…ng an individual ID from the context must happen before closing the applicable device. Also a workaround in spi testcases.  closing the spi device no longer removes the spi IDs from the context.


pi4j-test/src/main/java/com/pi4j/test/smoketest/SpiTestCase.java
```diff
@@ -54,6 +54,8 @@ public static TestResult run(ProviderContext providerContext) {
             return new TestResult(TEST_NAME, false, "Test failure: " + e.getMessage());
         } finally {
             if (spi != null) {
+                // this is a workaround as spi.close() no longer removes the ID  runtime context
+                providerContext.getContext().shutdown(spi.id());
```
FDelporte, line 58: is this something which shuld be fixed in the core where close is handled?

taartspi:
Yes. A core big. I fixed it with a different PR

PR #677, opened 2026-06-24
spi.close() does not remove the ID from the Context.  Not sure how th…

taartspi:
…e context rework created this, calling super.close() fixed it.  Maybe Stefan will understand the previous code path that did the removel.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/spi/SpiBase.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/spi/SpiBase.java
@@ -64,6 +64,7 @@ public abstract class SpiBase extends IOBase<Spi, SpiConfig, SpiProvider> implem
     @Override
     public void close() {
         logger.trace("invoked 'closed()'");
+        super.close();
         this.isOpen = false;
     }
 }


stefanhaustein (approved):

stefanhaustein:
(I suspect it never worked and cleaning up a supposedly redundant shutdown call somewhere exposed this)?

pi4j-core/src/main/java/com/pi4j/io/spi/SpiBase.java
```diff
@@ -64,6 +64,7 @@ public void open() {
     @Override
     public void close() {
         logger.trace("invoked 'closed()'");
+        super.close();
```
FDelporte, line 67: Based on #676, do we also need this for I2C?

stefanhaustein: IIRC I had fixed this a while ago for I2c but did not check SPI: 
https://github.com/Pi4J/pi4j/blob/7acbaa4b36e4dc6a96d713db24bcc441c77bda59/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBase.java#L69

I had noticed while working on the USB bridge, which supports I2C only...

I guess it might be cleaner to call super.close there, too

eitch:
@stefanhaustein will you change it in SPI then?

taartspi:
The spi smoke tests only used spi.close().  So after the context rework the
second spi test failed as the first one had not removed the ID.   I fixed
spi  with a call to super which is different than i2c was fixed.    I
believe Stefan prefers the super call i used. I can change the i2c to use
super

Issue #691, opened 2026-07-14
Pwm device initialisation inconsistency affects shutdown/close behaviour


IAmNickNack:
I have been investigating the changes required to get gRPC working with the latest `main` v5 snapshot.

The current gRPC implementation takes/took advantage of the mutable registry and the removal of the mutating methods meant a review of how the server disposes of devices in response to a `close()` request. 

The solution appears to be pretty simple, switching to `Context.shutdown(id)` or `IOBase.close()`, etc. In doing this however, I got breaking unit tests for PWM when asserting that devices had been removed from the server-side remote context.

Looking at the code for `PwmBase`, it appears that the `initialize(context)` override doesn't call `super.initialize(context)`. This means that `IOBase.context` isn't set, therefore `IOBase.close()` skips the `context.shutdown(id)` call and a PWM device does not get removed from the context.

Other device types either don't override `initialise` or make the call to `super` in their implementation and behave as expected when closed.

Assuming that the omission of the call to the base class from `PwmBase` is not intentional, I have a PR ready to raise to resolve what I currently think is an issue.

FDelporte:
Merged

PR #692, opened 2026-07-14
call super.initialize(context) in PwmBase

IAmNickNack:
PR to address #691


FDelporte:
Based on the changes you did in the unit test, this seems to make total sense! Thanks for researching and proposing the fix. @stefanhaustein can you also review please as this seems to be related also to the rework of `.close()` you did earlier?

stefanhaustein (approved):
Thanks!

taartspi (commented):
Ok

Issue #690, opened 2026-07-11
Can we kill PwmPresets?


stefanhaustein:
Just stumbled upon them when looking into PWM for PCA9685 support...

Can we remove them? They were missed in the conversion to double and nobody seems to have noticed so far....

We already have `on(double dutyCycle, double frequency)`, which seems to set both atomically and should enable users to implement presets in a straightforward manner.

EDIT: The alternative would be to adjust them to doubles, too.

PR #693, opened 2026-07-16
Delete PWM presets (#690)

stefanhaustein:


FDelporte:
I agree that the presets don't add much functionality as they only hold the two values that can be set directly. No idea, if they are in use...

FDelporte (approved):
OK for me, in normal release schedule I would think Deprecating first would be a better way forward, but as we are going to release as V5, direct removal is indeed an option.

stefanhaustein:
> OK for me, in normal release schedule I would think Deprecating first would be a better way forward, but as we are going to release as V5, direct removal is indeed an option.

- The reason I went for removal directly is that the presets no longer match the value types -- which were changed to doubles to improve precision. They might still work anyway but there are no test and it just didn't seem worth the hazzle
- They add complexity when implementing PWM in a PCA9685 driver
- Nobody objected to the suggestion to remove them so far, see #690

taartspi (commented):
I agree to remove them cleanly at V5.    And not sure how one would use them

Issue #695, opened 2026-07-17
Extend possibilities to get address/bcm per type of board


FDelporte:
Banana Pi Zero = the index of the bank (A=0, B=1, C=2, … I=8) multiplied by 32, plus the pin number within that bank
See https://webtechie.be/post/first-test-of-java-on-banana-pi-arm-and-risc-v-plus-a-blinking-led-with-pi4j/
First implementation proposal: https://github.com/Pi4J/pi4j/blob/main/pi4j-core/src/main/java/com/pi4j/boardinfo/util/LineHelper.java

Nvidia jetson nano uses Tegra GPIO = multiplying by 8 instead of 32
https://forums.developer.nvidia.com/t/tegra-gpio-port-offset/179621?utm_source=chatgpt.com

OrangePi uses Allwinner in some boards, same calculation + Rockchip rk3566 and others
https://wiki.t-firefly.com/en/ROC-RK3566-PC/driver_gpio.html?utm_source=chatgpt.com

stefanhaustein:
Probably should just not be static and take lines per bank as a ctor param to cover more chips?

What are the alternatives? Should we re-open a more straightforward IO building discussion?

If we had clean IO builders, we could potentially move the whole config stuff to a compatibility package that's backed by the builders?

DigitalSmile:
I think we should stick to linux specific terms (line / offset), at least internally. But for user we need to provide convenient methods to pass SBC specific labels (like "GPIO17", "PI15" and others stated in chip documentation) to our code, which should be mapped accordingly. So probably we can introduce some kind of interface:

```java
public interface PinAddress {
    record ByOffset(int chip, int offset)        implements PinAddress {}
    record ByName(String lineName)               implements PinAddress {}
    record ByHeaderPin(int physicalPin)          implements PinAddress {}
    record Bcm(int bcm)                          implements PinAddress {} // RPi legacy
}

ChipLine resolve(PinAddress a) {
    return switch (a) {                          // exhaustive pattern switch
        case ByOffset(var chip, var off) -> new ChipLine(chip, off);
        case ByName(var n)               -> discovered.getOrThrow(n);
        case ByHeaderPin(var p)          -> board.header().getOrThrow(p);
        case Bcm(var bcm)                -> new ChipLine(0, bcm); // Pi mapping
    };
}
```

[...]

PR #696, opened 2026-07-17
Added LineHelper for non-Raspberry Pi-boards

FDelporte:
Related to #695

Draft extending LineHelper to support more GPIO interfaces

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/boardinfo/util/LineHelper.java
+++ b/pi4j-core/src/main/java/com/pi4j/boardinfo/util/LineHelper.java
@@ -9,48 +9,89 @@ package com.pi4j.boardinfo.util;
  * banks behind a single {@code gpiochip} character device, so a pin such as {@code PI15} (bank I,
  * pin 15) must be translated to its absolute line offset ({@code 8 * 32 + 15 = 271}) before it can be
  * used with the GPIO character-device API.</p>
+ *
+ * <p>Other SoC families use a different number of lines per bank. NVIDIA Tegra SoCs, as used on
+ * boards like the Jetson Nano, group their GPIOs into lettered ports of 8 pins each instead of 32
+ * (e.g. {@code PZ.00}), so the same bank/line pair resolves to a different absolute offset there.
+ * Use the overloads accepting a {@link GpioLineFamily} to target a family other than the default
+ * {@link GpioLineFamily#ALLWINNER}.</p>
+ *
+ * @see GpioLineFamily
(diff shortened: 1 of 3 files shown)


pi4j-core/src/main/java/com/pi4j/boardinfo/util/LineHelper.java
```diff
@@ -9,48 +9,89 @@
...
-     * and its pin number within that bank.
+     * and its pin number within that bank, assuming the {@link GpioLineFamily#ALLWINNER} family.
      *
      * @param bank the bank letter ({@code 'A'}-{@code 'Z'}, case-insensitive), e.g. {@code 'I'} for PI15
      * @param line the pin number within the bank ({@code 0}-{@code 31}), e.g. {@code 15} for PI15
      * @return the absolute line offset on the {@code gpiochip} device, e.g. {@code 271} for PI15
-     * @throws IllegalArgumentException if {@code bank} is not a letter or {@code line} is not in {@code 0}-{@code 31}
+     * @throws IllegalArgumentException if {@code bank} is not a letter or {@code line} is out of range
      */
     public static int getAddress(char bank, int line) {
```
stefanhaustein, line 43: Are we sure the letters map the same for all chips?

pi4j-core/src/main/java/com/pi4j/boardinfo/util/LineHelper.java
```diff
@@ -9,48 +9,89 @@
...
      *
      * @param bank the zero-based bank index ({@code 0} for bank A, {@code 1} for bank B, ...), e.g. {@code 8} for bank I
      * @param line the pin number within the bank ({@code 0}-{@code 31}), e.g. {@code 15} for PI15
      * @return the absolute line offset on the {@code gpiochip} device, e.g. {@code 271} for PI15
-     * @throws IllegalArgumentException if {@code bank} is negative, {@code line} is not in {@code 0}-{@code 31},
+     * @throws IllegalArgumentException if {@code bank} is negative, {@code line} is out of range,
      *                                   or the resulting offset would overflow {@code int}
      */
     public static int getAddress(int bank, int line) {
+        return getAddress(GpioLineFamily.ALLWINNER, bank, line);
```
stefanhaustein, line 58: Why stick to static instead of handing in the count / chip to a ctor?

Isn't it likely that multiple gpios will be created for the same chip?

For backwards compat (if desired?) just deprecate the existing static methods?

Issue #26, opened 2020-07-06
Multipurpose (Input & Output) GPIO


treitmaier:
**The Context**
I'm in the process of porting the libraries of [Skywriter](https://github.com/pimoroni/skywriter-hat) & [Flick](https://github.com/PiSupply/Flick) I2C gesture sensing boards from python to Java/pi4j. Both boards utilise the MGC3130 chipset.

**The Problem**
The MGC3130 chipset requires the transfer pin (BCM 27) to be used as both input and output within each polling cycle.  At the start of the cycle the library waits for a low transfer line (*input*). However, the library then reconfigures the same pin (BCM27) as a low *output* to assert that the MGC3130 does not update data buffers before reading from them.

**The Issue**
Does pi4j-v2 support Multipurpose GPIO that can be reconfigured to work as input *and* output?

FDelporte:
Thanks for your question! And nice to see you are working on these libraries using Pi4J!

As you can see in https://github.com/Pi4J/pi4j-v2/tree/master/pi4j-core/src/main/java/com/pi4j/io/gpio/digital currently we only have input or output types. Would it be an idea to reconfigure the I/O in your library itself between each action? I've not done this myself before so no idea what the behavior would be and if this happens fast enough.

FYI, under the hood, Pi4J V2 is using http://abyz.me.uk/rpi/pigpio/ so anything supported by that library should also be available in Pi4J or could be added.

savageautomate:
I remember something like this in Pi4J V1.  
See:  https://github.com/Pi4J/pi4j/search?q=GpioPinDigitalMultipurpose&unscoped_q=GpioPinDigitalMultipurpose

We probably should consider this for V2 as well.

treitmaier:
Thanks for your comments @savageautomate & @FDelporte . It looks like pigpio would support Multipurpose GPIOs through its [gpioGetMode](http://abyz.me.uk/rpi/pigpio/cif.html#gpioGetMode) and [gpioSetMode](http://abyz.me.uk/rpi/pigpio/cif.html#gpioSetMode) methods.

[...]

PR #698, opened 2026-07-20
Refactoring of GPIO FFM

DigitalSmile:
1) Removes code duplication
2) Added an ability to reconfigure the pin at runtime (https://github.com/Pi4J/pi4j/issues/26)

All unit and integration tested.


plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMGpioLine.java
```diff
@@ -0,0 +1,159 @@
...
+ * via {@link #writeState}, and released via {@link #close}.
+ */
+class FFMGpioLine {
+    private static final Logger logger = LoggerFactory.getLogger(FFMGpioLine.class);
+
+    final IoctlNative ioctl = new IoctlNative();
+    final FileDescriptorNative file = new FileDescriptorNative();
+
+    final String deviceName;
+    final int bcm;
```
eitch: are we actually still talking BCM here?

plugins/pi4j-plugin-ffm/src/test/resources/gpio-setup.sh
```diff
@@ -8,7 +8,7 @@
 #   GPIO_MOCK_HOGS    - occupied line per chip,   (default "2,-1", -1 means none)
 #   GPIO_MOCK_NUMBERS - desired /dev/gpiochip<N>  (optional, e.g. "0,1,99")
 # Example: GPIO_MOCK_NGPIOS=16,4 GPIO_MOCK_LABELS=bankA,bankB GPIO_MOCK_HOGS=-1,-1 ./gpio-setup.sh
-NGPIOS="${GPIO_MOCK_NGPIOS:-8,1}"
+NGPIOS="${GPIO_MOCK_NGPIOS:-9,1}"
```
eitch, line 11: What's this change?

eitch (commented):
So basically i like the refactoring to use the GpioLine. A little sad we are adding the reconfigure in the same PR, but the code looks good.

Just wondering if a reconfigure interface is the right call, but didn't come up with a better idea while reviewing the code, so i guess we can merge. But let's also see what the others say.

stefanhaustein:
I second Robert's opinion: 
- Ideally, we'd have a separation of concerns here
- Let's not expand the public API without an issue / proper API discussion.

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMGpioLine.java
DigitalSmile: I think we can change it to offset.

plugins/pi4j-plugin-ffm/src/test/resources/gpio-setup.sh
DigitalSmile: It is changing mock driver configuration - adding more available offsets to test runner.

stefanhaustein:
> Just wondering if a reconfigure interface is the right call, but didn't come up with a better idea while reviewing the code, so i guess we can merge. But let's also see what the others say.

For instance, we could open a "bidirectional" pin in the configuration phase that would implement both DigitalInput and DigitalOutput and then the change of direction could be implied by whether the value is set or queried. 

Another option would be an explicit bidirectional interface.

I am not saying "this is the way", but we should iterate all the options and then take it from there.

stefanhaustein (changes requested):
(see top level comments)

eitch:
> For instance, we could open a "bidirectional" pin in the configuration phase that would implement both DigitalInput and DigitalOutput and then the change of direction could be implied by whether the value is set or queried.
> 
> Another option would be an explicit bidirectional interface.

Yeah, i was also thinking, that it would make it more sense to not need to reconfigure so "heavily". Perhaps we can have a class DigitalInputOutput, having the meaning in the name.

In another discussion there was the though of removing "Digital" - But is then GpInputOutput better? Only having Input, Output and InputOutput seems not quite clear.

Or @stefanhaustein would you want to call it BidirectionalIO? That seems strange too.

stefanhaustein:
I'd be fine with DigitalInputOutput or DigitalBidirectionalIo or BidirectionalIo -- I think the important part is to have a dedicated interface as it seems like a dedicated use case that typically needs special support from the connected device. 

I am not a big fan of removing "Digital" from Input/Output, as there are probably boards out there that support Analog IO directly, and it seems like we are on a path to supporting more boards.

EDIT: AI answer for sbc with analog built-in: https://share.google/aimode/64DcG8QDveYYhYcRI

DigitalSmile:
I have removed all reconfigure things, left just GPIO refactoring,

> For instance, we could open a "bidirectional" pin in the configuration phase that would implement both DigitalInput and DigitalOutput and then the change of direction could be implied by whether the value is set or queried.
> Another option would be an explicit bidirectional interface.
> I am not saying "this is the way", but we should iterate all the options and then take it from there.

I did not make any interface because it may confuse the user. "Bidirectional" means it can simultaneously write/read from pin, which is not true at all. I don't think there should be any "WhateverInputOutput"...

> In another discussion there was the though of removing "Digital" - But is then GpInputOutput better? Only having Input, Output and InputOutput seems not quite clear.

We removed "Analog" from out code because it is not working the same way as GPIO. It is different subsystem in kernel (DAC/ADC or IIO) and there should be different architectural approach. In terms of GPIO there is no "Digital" pin, every pin essentially IS digital.

> EDIT: AI answer for sbc with analog built-in: https://share.google/aimode/64DcG8QDveYYhYcRI

Again, that's not abstraction model we have in gpio package of core, it works completely different.

---

I have created two separate issues to discuss: 
- https://github.com/Pi4J/pi4j/issues/699
- https://github.com/Pi4J/pi4j/issues/700

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/EventWatcher.java
```diff
@@ -0,0 +1,175 @@
...
+    static final int EVENT_WATCHER_SHUTDOWN_TIMEOUT_MS = 200;
+
+    private final int fd;
+    private final int offset;
+    private final FileDescriptorNative file;
+    private final PinEvent pinEvent;
+    private final PinEventProcessing eventProcessor;
+    private final long debounceNs;
+
+    private boolean stopWatching = false;
```
eitch: I think this should be volatile

eitch (approved):
looks good with exception of the volatile flag fro running

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/EventWatcher.java
DigitalSmile: Yes, missed that.

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutput.java
```diff
@@ -52,7 +52,6 @@ static DigitalOutputBuilder newBuilder(Context context){
         return DigitalOutputBuilder.newInstance(context);
     }
 
-
```
stefanhaustein, line 55: Please avoid pulling in files that are not "really" changed

eitch: I saw this too, but ignored it, as it was from a reverting of the other changes, thus i merged

PR #702, opened 2026-07-25
Remove providers and adjust the code accordingly

stefanhaustein:


stefanhaustein:
not proceeding at this time due to ProviderProvider shortcuts being documented

PR #711, opened 2026-07-26
Nws/remove generics from config types

IAmNickNack:
This PR removes unnecessary type parameters from configuration classes, aiming to simplify the code in this area. 

It includes no functional changes. However, it does include breaking changes to the public API. It does not include any changes to the builder APIs.

Relates to #705

PR #716, opened 2026-07-28
declare generic type arguments in builder interfaces

IAmNickNack:
This PR takes advantage of #711 to clarify the generic type arguments used on builder types.

Functionally, this doesn't change anything. As with anything generics-related, it's largely cosmetic, but it does stop and IDE's linter from warning about unparameterised type declarations.

This does emphasise the complexity that can that is perceived to exist in the builder hierarchy


IAmNickNack:
I don't agree with the Sonarcube assessment regarding duplication. The only way to remove the duplication would be to remove generics

IAmNickNack:
Also, the maintainability assessment could be argued for the same code prior to modification

FDelporte (approved):

Issue #704, opened 2026-07-26
What is the canonical way of creating an IO instance?


stefanhaustein:
On this [page](https://www.pi4j.com/documentation/build-io/), we have:

1. `var button = pi4j.din().create(INTEGER_PIN_ADDRESS);`
2. `var button = pi4j.digitalInput().create(INTEGER_PIN_ADDRESS);`
3. 
```
var buttonConfig = DigitalInput.newConfigBuilder(pi4j)
        .id("button")
        .name("Press button")
        .address(PIN_BUTTON)
        .pull(PullResistance.PULL_DOWN)
        .debounce(3000L);
var button = pi4j.create(buttonConfig);
```

I presume `var button = pi4.din().create(buttonConfig)` and `var button = pi4.digitalInput().create(buttonConfig)` would also work....?

In the code, I can additionally find  `DigitalInputConfig.newBuilder()` and `DigitalInputConfigBuilder.newInstance()`

I have some questions about this:

1. What is the canonical way? I suppose `DigitalInput.newConfigBuilder` but would appreciate an explicit confirmation?
2. Why do we publicly document other ways? Do we want to preserve the ProviderProvider shortcuts? And extend them to new config options? Or should we phase them out?
3. Is address still a thing or should this be bcm?
4. Can we deprecate everything non-canonical?
Alternative ways of doing the same thing make it hard to read and maintain code.

IAmNickNack:
> Is address still a thing or should this be bcm?

`address` is deprecated. 

This could be removed from the actual configs now and builders could treat `address` as an alias for `bcm`, keeping the deprecation warning on the builder class for now. (?)

IAmNickNack:
A canonical create/startup would also simplify shutdown logic. Possibly negating the need for a `shutdownInternal` operation

IAmNickNack:
> What is the canonical way? I suppose DigitalInput.newConfigBuilder but would appreciate an explicit confirmation?

This would be my assumption

> Do we want to preserve the ProviderProvider shortcuts? And extend them to new config options? Or should we phase them out?

If the canonical method is as you suggest, providing API access to individual providers seems like an anti-pattern as it allows the user to bypass or short-circuit significant parts of lifecycle management. 

> Can we deprecate everything non-canonical? Alternative ways of doing the same thing make it hard to read and maintain code.

Absolutely. The ambiguities can make simple changes unnecessarily difficult to make and validate.

IAmNickNack:
[...]

PR #721, opened 2026-07-30
fix: correct SPI IOType enum entry

IAmNickNack:
`SPI` incorrectly specified i2c config and config builder types.

This would have caused obvious problems if it was ever used. 

It would be invoked by `Provider.create(String)`. So It might be fair to assume that this function is never used and can be removed. (Also, it doesn't seem to make sense to create a device with only and id and no configuration)

https://github.com/Pi4J/pi4j/blob/2fdff679ee9579c8103d2b30d22b88c4022b7e21/pi4j-core/src/main/java/com/pi4j/provider/Provider.java#L96-L104

invokes:

https://github.com/Pi4J/pi4j/blob/77be80407483c5ad9a7680750fa31fe6bdc108f9/pi4j-core/src/main/java/com/pi4j/io/IOType.java#L103-L110

In the latter, I'd question why runtime reflection is required and suggest code-smell. It's probably doesn't participate in any canonical construction workflow and should also be removed

#704

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/IOType.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/IOType.java
@@ -3,6 +3,7 @@ package com.pi4j.io;
 import com.pi4j.context.Context;
 import com.pi4j.exception.Pi4JException;
 import com.pi4j.io.gpio.digital.*;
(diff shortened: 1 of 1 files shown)


FDelporte (approved):
good catch!

PR #722, opened 2026-07-30
chore: remove references to platform

IAmNickNack:
If platforms are no longer a thing, references to platform can be removed.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/context/ContextBuilder.java
+++ b/pi4j-core/src/main/java/com/pi4j/context/ContextBuilder.java
@@ -27,9 +27,6 @@ public class ContextBuilder implements Builder<Context> {
     protected boolean autoDetectProviders = false;
     protected boolean enableShutdownHook = false;
 
-    // default platform identifier
-    protected String defaultPlatformId = null;
-
     // extensibility modules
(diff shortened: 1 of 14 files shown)


FDelporte:
This makes me wonder, should references to "provider" be renamed to "plugin"?

eitch (approved):

IAmNickNack:
Having removed `platform`, I'm still looking at config related stuff. The only purpose left for `IOConfig` now is to expose a `provider` property. (example hierarchy for `DigitalInputConfig` below).

With #704 in mind (canonical device creation), I wonder whether `provider` needs to exist in Config anywhere. And if not, `IOConfig` becomes pretty redundant other than becoming a marker class. 

If the context only has a single device plugin (ffm, mock, etc) registered now and a single provider per device type , there's no need to lookup a specific provider by name. When a config is provided to an `IOCreator`, because of `IOType`, the creator knows which provider maps to which config type. There's no ambiguity, so no name lookup is needed (I've not had a need to use provider names myself in the past).

I suppose there's an edge case where the user adds two different providers for the same `IOType`, but I can't imagine a valid use-case for this. If the user really want's to use two different providers for the same device type, they can use a second context.

image

PR #726, opened 2026-08-02
call super after custom shutdownInternal implementation in I2CBase

IAmNickNack:
Calling `close()` from `shutdownInternal()` doesn't seem right. `close()` eventually invokes `shutdown(id)` which in turn calls `shutdownInternal()`. This is possibly unnecessary complexity. If such logic is required, it could suggest a more fundamental problem.

This change makes `I2CBase` behaviour more consistent with other devices and also allows the base class method to be invoked.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBase.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/i2c/I2CBase.java
@@ -1,7 +1,5 @@
 package com.pi4j.io.i2c;
 
-import com.pi4j.context.Context;
-import com.pi4j.exception.ShutdownException;
 import com.pi4j.io.IOBase;
 import com.pi4j.io.i2c.impl.DefaultI2CRegister;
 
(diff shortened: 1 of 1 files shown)


IAmNickNack:
I am assuming `RegistryTest#testCreateMultipleSameAddress()` validates this change

IAmNickNack:
Without the call to `close()`, FFM tests appear to fail.  Does this suggest this logic is in place to solve a possible issue in FFM?

I've updated the change to add back the call to `close()` but have kept the call to the base method. 

The test which failed did not execute locally on MacOS

stefanhaustein:
The close / shutdown dance is documented here:

https://github.com/Pi4J/pi4j/blob/a367cfdf26e1ce0acea7e65d0132c09d68cfb47b/pi4j-core/src/main/java/com/pi4j/io/IOBase.java#L66

The reason this is a bit convoluted is that we had to retro-fit `close()`. 

We might want to make `close()` final so people don't override it. This should make it impossible that io subclasses distribute functionality between `close()` and `shutdownInternal()` in weird ways.

And then perhaps rename `shutdownInternal()` to `closeImpl()` or `closeInternal()` or similar?

eitch:
Do you guys want to merge this, or refactor the name and make the method final?

IAmNickNack:
My intention was to merge this as a fix for the existing code.

I think there are other initiatives focusing on construction and lifecycle which could address the confusion around shutdown and close.

stefanhaustein:
> My intention was to merge this as a fix for the existing code.
> 
> I think there are other initiatives focusing on construction and lifecycle which could address the confusion around shutdown and close.

I have created #736 to track the close cleanup

mstahv:
FYI: faced this same issue recently, thanks for the fix 🙏

PR #727, opened 2026-08-03
Feat/prep FFM support multiple GPIO lines

IAmNickNack:
This PR does some prep work for adding a parallel port device to Pi4J. 

It swaps out a single `offset` integer value for either an array of offsets or a bitmask as required by the `ioctl` calls. 

This doesn't appear to negatively affect existing behaviour, but I am unable to run GPIO smoke tests to check fully.

A majority of the changes in this PR affect logging and how the pin mask is represented.

This is a small but significant change. It could probably be merged as-is, or we could stack additional PRs on top of it until the parallel implementation arrives.

Finally, it updates the Checkstyle plugin to use the latest engine. Older versions don't support java 25 construction when operations are performed before a call to `this()`.


IAmNickNack:
I'm not sure what the Sonar report is complaining about. It doesn't give any detail.

It mentions branches. I'm wondering if this is the nested `if`s in the `EventWatcher`

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMGpioLineMask.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.List;
+
+/**
+ * Utility record which can be used to represent both the mask for a set of line offsets and the offsets for
+ * a mask (whichever is already known)
+ * @param offsets The offsets of the GPIO lines.
+ * @param mask The mask representing the GPIO lines.
+ */
+public record FFMGpioLineMask(
+    int[] offsets,
```
stefanhaustein: Wouldn't storing the mask be sufficient? 

You could provide an offset iterator here and make it skip empty bytes or something to speed up if you are concerned about performance

Or maybe drop this whole class and do this statically? 

Or just use a BitSet: https://stackoverflow.com/a/15393089/1401879

IAmNickNack: ~~You're right to call this out in this PR. It's not really relevant. I'll tidy this up.~~

I have a stack of PRs on my fork which phase the implementation of a parallel port. The concept does become more relevant later, but it would be more appropriate to discuss it in the correct context and leave it out of this PR.

Edit: 

I'm just reviewing the use of this and I'm going to stand by it for now. In the current instance, it simplifies iteration and logging used in the in-use check and error reporting.

IAmNickNack: > Or just use a BitSet: https://stackoverflow.com/a/15393089/1401879

I've never been convinced by `BitSet`. These types of operation are arguably more straightforward using numeric bit masks.

stefanhaustein: Won't a record provide an implicit getter for offsets? And then one could manipulate offsets, leading to an inconsistent state? What if somebody changes offsets after handing them in? 

Storing the same information twice and somehow the order is only half relevant (iteration but not equals) kind of creeps me out. And then the offsets array could have the same index multiple times. What would that mean? Will it be visited multiple times? 

IMHO, this should just be mask. And then it feels odd to have a class around it.

Or a Set<Integer>. And then again producing the mask could just be a static function. 

Or just a bitset, which is backed by numeric bit masks, so I don't really understand your concern about that? https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/util/BitSet.java#L90

I mean if there is some higher purpose in storing a bit set other than as a bit set it should be straightforward to explain now?

IAmNickNack: I take your point about mutability

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/MaskUtils.java
```diff
@@ -0,0 +1,69 @@
...
+            this.remaining = mask;
+        }
+
+        @Override
+        public boolean hasNext() {
+            return remaining != 0;
+        }
+
+        @Override
+        public Integer next() {
```
stefanhaustein, line 59: Consider adding int nextInt() to avoid boxing

stefanhaustein (approved):

plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/MaskUtils.java
IAmNickNack: Then it's not an Iterator is it?

This is not going to be used at runtime during IO. Boxing and performance are not really a concern at that point.

If a user is reconfiguring at runtime, they're unlikely to be concerned with microsecond optimisations

IAmNickNack: If it were code in a hot-path, maybe it would be something worth worrying about. But it's also amazing what HotSpot does these days. I'd be surprised if this kind of thing didn't get optimised out (after a few 100,000 iterations)

stefanhaustein: > Then it's not an Iterator is it?

The iterator method could just call nextInt, but it was just a suggestion and I think numbers below 128 are allocation free IIRC anyway....

stefanhaustein:
I am ignoring the code analysis for now as it provides an upsell link instead of details and they are probably pre-existing.

PR #733, opened 2026-08-10
remove calls to deprecated factory methods

IAmNickNack:
This PR removes calls to factory methods deprecated in #728

`ProviderProvider` is deprecated. Factory methods it declares are deprecated too


stefanhaustein (approved):
Thanks, LG

PR #734, opened 2026-08-11
register devices during `context.create(config)`

IAmNickNack:
This PR relates to the discussion on canonical construction in #704 and is encouraged by the deprecation of `ProviderProvider` in #728 

It assumes a minimal number of construction paths and centralises functionality that might be expected of the context as a result of constructing via `context.create(config)`. 

* Devices are registered by the context during `create`. The context now owns all the mutation during device construction transaction and is responsible for maintaining its own state.
* Providers are no longer required to the implement boilerplate `context.register(device)`. `Provider.create(...)` is now "pure", with the single responsibility to construct the device.

Based on suggestions [in this comment](https://github.com/Pi4J/pi4j/issues/704#issuecomment-5242513000), it aims to give the two affected execution paths clearer intention.

--- 

I'm tempted to suggest taking this further by removing/deprecating `Context.register(IO)`:

* This PR would mean it's no longer called from anywhere in the main codebase. (we don't need it).
* It would remove some ambiguity around construction responsibilities. Users wouldn't have the option of registering devices outside a "canonical" execution path.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/context/Context.java
+++ b/pi4j-core/src/main/java/com/pi4j/context/Context.java
@@ -222,14 +222,16 @@ public interface Context extends Describable, IOCreator, ProviderProvider, Initi
     // I/O INSTANCE ACCESSOR/CREATOR METHODS
     // ------------------------------------------------------------------------
 
+    @SuppressWarnings({"unchecked", "rawtypes"})
     @Override
     default <I extends IO<?, ?, ?>> I create(IOConfig config, IOType ioType) {
+        I device = null;
         // create by explicitly configured IO <PROVIDER> from IO config
         String providerId = config.provider();
         if (StringUtil.isNotNullOrEmpty(providerId)) {
             // resolve the provider and use it to create the IO instance
             Provider provider = this.providers().get(providerId, ioType);
-            return (I) provider.create(config);
+            device = (I) provider.create(config);
         }
 
         // get implicitly defined provider (defined by IO type)
@@ -237,7 +239,12 @@ public interface Context extends Describable, IOCreator, ProviderProvider, Initi
         if (ioType != null) {
             // resolve the provider and use it to create the IO instance
             Provider provider = this.provider(ioType);
-            return (I) provider.create(config);
+            device = (I) provider.create(config);
+        }
+
+        if (device != null) {
+            register(device);
+            return device;
         }
 
         // unable to resolve the IO type and thus unable to create I/O instance
--- a/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMDigitalInputProviderImpl.java
+++ b/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/gpio/FFMDigitalInputProviderImpl.java
@@ -1,8 +1,5 @@
 package com.pi4j.plugin.ffm.providers.gpio;
 
-import com.pi4j.context.Context;
-import com.pi4j.exception.InitializeException;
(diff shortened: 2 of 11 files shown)


IAmNickNack:
Build is currently failing. I expect this is due to what is resolved by #735

eitch:
Can you rebase with develop and try again?

IAmNickNack:
> Can you rebase with develop and try again?

🤞

eitch (approved):
looks good

eitch:
Are all happy with these changes? @FDelporte @stefanhaustein

stefanhaustein (approved):
lgtm

stefanhaustein:
> I'm tempted to suggest taking this further by removing/deprecating `Context.register(IO)`:

I am fine with this as long as we keep this exposed to subclasses in DefaultContext (= have protected visibility there or  make the mutable registry accessible to subclasses)

Wrt deprecation vs. removal, I think for v5 we should be fine to remove what we don't expect clients to call but we should deprecate what users might (still) be using. For instance, some outdated construction methods are still advertised on our website.

IAmNickNack:
> Wrt deprecation vs. removal, I think for v5 we should be fine to remove what we don't expect clients to call but we should deprecate what users might (still) be using. For instance, some outdated construction methods are still advertised on our website.

I'm only thinking out loud, but one possible solution to this would be dynamic proxies which provide the existing behaviour for deprecated methods, but use the POD implementations that the maintainers are more interested in building. 

Under the covers we build the thing we want to build, without the constraints of accommodating deprecation and non-canonical execution paths. We **do** have to model and build the proxies, but this is fun! On the surface, users get deprecation warnings during compilation, but no runtime penalty. 

Users get deprecation, Maintainers get to do useful work. Everyone's a winner! 

Like I said, I'm only thinking out loud. 

--- 

Theoretically, you could fork the repo and implement all the existing interfaces in this way for backwards compatibility. Behind the scenes it could be an entirely different implementation.

IAmNickNack:
If the interfaces conveying the object graph for the next version POD exist before any real implementation work starts, the first development task could be then to build a `shim` implementation for the design which confirms to the existing design. 

Once this exists, the next-gen API is useable and provable and the design is confirmed by achieving compatibility with the current version. 

Implementation of next-gen and phase-out of the current API are then detached. The next-gen implementation is unconstrained by current implementation and free to do what it wants so long as the shim can still function.

I think it's possible that this a similar approach to that taken the by Open Telemetry project during the shift away from Open Tracing. They built a shim which provided the existing API on top of an implementation of the next-gen. 

Maybe I live some kind of utopia, where this works seamlessly :D 

--- 

Edit: This might be an alternative approach to chipping away at v1/v2

stefanhaustein:
I know I like to complain about aspects of the API, but in the end it gets the job done. From a user perspective, a new API is probably an annoyance -- I'd dread converting drivers to a new API. And documentation and examples showing the "old" way will probably stay on the web forever, confusing users. So I think I'd prefer to stay on the current course of incremental improvements and cleanups, even if a clean new API while preserving the old behaviour is clearly possible via proxying.

stefanhaustein:
PS.: I think I have overlooked one problem here: Doesn't this mean IO instances created with the deprecated methods will no longer be registered and hence no longer work properly?

IMHO Deprecation is worse than removal where it breaks compatibility at runtime, basically turning what would be a compile time issue (easy to spot) into a runtime problems (debugging fun with no upper time bound)

IAmNickNack:
Good catch. I need to go to work, but happy to propose something to solve this

IAmNickNack:
https://github.com/Pi4J/pi4j/pull/738 can resolve the execution paths broken by this change

Issue #736, opened 2026-08-11
Clarify / clean up the IO shutdown sequence


stefanhaustein:
- (at least) make close final in IOBase
- Remove existing calls to close from shutdowninternal (perhaps implement some kind of guard against this pattern if feasible)

cf https://github.com/Pi4J/pi4j/pull/726#issuecomment-5244386611

PR #737, opened 2026-08-11
Make IOBase.close() final

stefanhaustein:
This forces IOBase to not implement close. Looking up the documentation for close then will point to shutdownInternal 

Clean up the fallout

- Add isOpen to the interface so we don't have it only in I2C where it would need to access the private closed state (widening visibility would make it writeable from elsewhere
- Remove some overrides that were doing nothing or redundant checks
- Move actual code over to shutdownInternal where it was misplaced

Background: #736


eitch (approved):
I gather you verified this, for me it looks good.

PR #738, opened 2026-08-12
Fix/support deprecated factory methods

IAmNickNack:
This PR allows deprecated `ProviderProvider` factories in the `DefaultContext` to remain functional after #734 likely caused them to stop working. (as pointed out [here](https://github.com/Pi4J/pi4j/pull/734#issuecomment-5258614740))

When these methods are called on the default context, the method now returns a shim implementation which calls the non-deprecated, favoured method of construction.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/context/impl/DefaultContext.java
+++ b/pi4j-core/src/main/java/com/pi4j/context/impl/DefaultContext.java
@@ -351,6 +351,7 @@ public class DefaultContext implements Context {
      * Provide backwards compatibility for deprecated {@link ProviderProvider#digitalInput()}
      */
     @SuppressWarnings("unchecked")
+    @Override
     public <T extends DigitalInputProvider> T digitalInput() {
         return (T) new DigitalInputProviderBase() {
             @Override
@@ -364,6 +365,7 @@ public class DefaultContext implements Context {
      * Provide backwards compatibility for deprecated {@link ProviderProvider#digitalOutput()}
      */
     @SuppressWarnings("unchecked")
+    @Override
     public <T extends DigitalOutputProvider> T digitalOutput() {
         return (T) new DigitalOutputProviderBase() {
             @Override
@@ -377,6 +379,7 @@ public class DefaultContext implements Context {
      * Provide backwards compatibility for deprecated {@link ProviderProvider#pwm()}
      */
     @SuppressWarnings("unchecked")
+    @Override
     public <T extends PwmProvider> T pwm() {
         return (T) new PwmProviderBase() {
             @Override
@@ -390,6 +393,7 @@ public class DefaultContext implements Context {
      * Provide backwards compatibility for deprecated {@link ProviderProvider#i2c()}
      */
     @SuppressWarnings("unchecked")
+    @Override
     public <T extends I2CProvider> T i2c() {
         return (T) new I2CProviderBase() {
             @Override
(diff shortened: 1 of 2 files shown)


IAmNickNack:
I'm not sure the quality gate is failing due to these changes. 

Given it's claiming something about branches (but won't tell us what), I would suspect it could be due to nested branching elsewhere in the file.

---

Edit: 

Oh. Well that rule had already been broken by the base class and would also be broken by all the concrete provider classes.

image


I think the root cause is elsewhere: 

image

`DigitalInputProviderBase`, `DigitalProviderBase` and `GpioProviderBase` don't actually do anything but declare constructors. Maybe these aren't necessary

IAmNickNack:
this would avoid the warning:

```java
    @SuppressWarnings("unchecked")
    public <T extends DigitalInputProvider> T digitalInput() {

        class DigitalInputProviderShim 
            extends ProviderBase<DigitalInputProvider, DigitalInput, DigitalInputConfig> 
            implements DigitalInputProvider {
            @Override
            public DigitalInput create(DigitalInputConfig config) {
                return DefaultContext.this.create(config);
            }
        }
        
        return (T) new DigitalInputProviderShim();
    }

```

... or we could remove the intermediate classes and redefine the actual providers in a similar manner to `DigitalInputProviderShim`

pi4j-core/src/main/java/com/pi4j/context/impl/DefaultContext.java
```diff
@@ -254,19 +275,19 @@ public Context shutdown() throws ShutdownException {
         this.shutdownEventManager.clear();
 
         return this;
-	}
```
stefanhaustein, line 257: please try to avoid noise that makes it harder to see the actual changes

pi4j-core/src/main/java/com/pi4j/context/impl/DefaultContext.java
```diff
@@ -325,4 +346,74 @@ public Context removeListener(InitializedListener... listener) {
...
         mutableRegistry.register(instance);
     }
+
+    /**
+     * Provide backwards compatibility for deprecated {@link ProviderProvider#digitalInput()}
+     */
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T extends DigitalInputProvider> T digitalInput() {
+        return (T) new DigitalInputProviderBase() {
```
stefanhaustein, line 356: I think it would be possible to use a single helper function if you use the create call with the explicit IO type

stefanhaustein (approved):

stefanhaustein:
> this would avoid the warning:

Let's not make things worse to make an automated test happy that was already broken; I think we can just override this

pi4j-core/src/main/java/com/pi4j/context/impl/DefaultContext.java
IAmNickNack: I don't think so. These are the function definitions from the deprecated interface. They all need to be implemented to override them

stefanhaustein: Yes sorry I thought  you could have  
```
 public <T extends DigitalInputProvider> T digitalInput() {
        return (T) myInternalCreateHelper(IOType.DIGITAL_INPUT)
```

BUT I overlooked that it needs to conform to the right generic type, so it's not really feasible

Issue #637, opened 2026-06-03
Do we still need the provider indirection at all in V5?


stefanhaustein:
I think I can experiment with implementing a context without providers for the MCP2221 Usb bridge, and if that works out, it might be worth trying the same for mainline / FFM....

stefanhaustein:
Still a bit crude but I think this should be workable:

https://github.com/stefanhaustein/pi4j-usb-bridge/blob/main/src/main/java/com/pi4j/usbbridge/DirectContextBase.java

eitch:
That makes sense

stefanhaustein:
@IAmNickNack would it be feasible to refactor your "remote Pi" work using a similar approach, dropping providers?

IAmNickNack:
Maybe. What's your beef with the current implementation? 

I'd agree that the current provider implantation can be problematic, with priority handling and service loader dependency. Maybe that's an unnecessary complexity. (I can expand on this point when I'm not on mobile).

From a broader design point of view, if we were to consider them factories they make a lot of sense and fit with the S in SOLID. Their single responsibility is to manage the lifecycle of the component at startup and shutdown and allow the device component to focus on the actual doing of the work at runtime. 

It's possible to move the lifecycle stuff elsewhere such as, as you suggest, the registry. But that clouds the registry's responsibility.

I'd say a big yes to removing the priority stuff and maybe also the service loader, but I'd question removing the abstraction for device specific lifecycle management.

stefanhaustein:
> Maybe. What's your beef with the current implementation?

It just looks like the opposite of KISS... Overuse of generics in too many places, reinventing a Map and calling it "Metadata",  fluff Interfaces without any clear purpose such as "Digital" all over the place, unclear responsibilities paired with overly-generic class names. What is a "Descriptor" and why does it have "parent" and "quantity"?

I am not complaining -- it just looks quit a bit grown. And complexity is really hard to refactor for new challenges. So if the loss of the provider requirement presents an opportunity to cut back radically, I think we should take it. 

> From a broader design point of view, if we were to consider them factories they make a lot of sense and fit with the S in SOLID. Their single responsibility is to manage the lifecycle of the component at startup and shutdown and allow the device component to focus on the actual doing of the work at runtime.

I was assuming the providers were the factories whereas the lifecycle was managed by some combination of the Context, (Runtime-?)Registry and Runtime? 

[...]

PR #740, opened 2026-08-15
Feat/add port interfaces

IAmNickNack:
This PR proposes a `ParallelPort` device to provide a superset of existing `Digital` devices, with two key differences:

* Multiple GPIOs can be read or written simultaneously (dependent on the configured direction)
* Direction can be explicitly reconfigured at runtime. 

## Multiple GPIOs

Simultaneous control of multiple GPIOs allows:

### Bus participation

Multiple GPIOs are represented as a bit mask and interpreted as `int`. This simplifies participating on a data bus. The complete value of a bus can be read or written in a single, atomic operation. This is preferable to to individually invoking multiple `Digital` devices to achieve the same effect.

### Virtual `Digital` devices

Once a `ParallelPort` is created, the user is still able to manipulate individual pins by writing the appropriate mask value to the port. It might be preferable however, to use the existing `Digital` abstractions to do this. 

A `ParallelPort` can be used as the backing device for `VirtualDigitalInputProvider` or `VirtualDigitalOutputProvider`. These are called out as being "virtual" because a) they don't directly control the hardware and b) their pin numbering is relative the the bit of the port they control and, as such are numbered sequentially  from `0`. 

For example, a 3-bit port to control a 3-to-8 decoder might be mapped to BCM pins 16, 20 and 21. This could be efficiently controlled by writing numeric values from 0 to 7. But, if the user wished to, they could create virtual digital devices on top of the port, where the pins would be numbered 0, 1 and 2. (This example might be unrealistic, but it illustrates the numbering scheme).

A more realistic use could be to for this mechanism to provide all digital devices. A port with a single BCM is essentially a digital device. A abstraction around the port can provide required `Digital.state` operations.

## Direction control

E.g. A `ParallelPort` could be used to participate in a data bus. Such a port could be configured with an initial direction of `INPUT`, allowing other participants control of the bus. When data is requested, the port might be required to write data and can be temporarily reconfigured as OUTPUT to do so.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/gpio/parallel/VirtualDigitalInputProvider.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/gpio/parallel/VirtualDigitalInputProvider.java
@@ -1,0 +1,127 @@
+package com.pi4j.io.gpio.parallel;
+
+import com.pi4j.io.gpio.digital.DigitalBase;
+import com.pi4j.io.gpio.digital.DigitalInput;
+import com.pi4j.io.gpio.digital.DigitalInputConfig;
+import com.pi4j.io.gpio.digital.DigitalInputProvider;
+import com.pi4j.io.gpio.digital.DigitalState;
+import com.pi4j.io.gpio.digital.DigitalStateChangeEvent;
+import com.pi4j.io.gpio.digital.DigitalStateChangeListener;
+import com.pi4j.provider.ProviderBase;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A provider for creating digital input instances from this parallel port.
+ * <p>
+ * The bcm numbering of digital devices created by this provider corresponds to the bits in the parallel port's value.
+ * E.g. If the {@link ParallelPort} is mapped to pins 10 and 12, the logical bcm of for the {@link DigitalInput}s
+ * creatable via this instance would be `0` and `1` respectively.
+ * <p>
+ * All {@link DigitalInput}s created by this provider share the same {@link ParallelPort} instance and therefore
+ * also the same event source.
+ */
+public class VirtualDigitalInputProvider
+    extends ProviderBase<DigitalInputProvider, DigitalInput, DigitalInputConfig>
+    implements DigitalInputProvider {
+
+    /**
+     * The parallel port instance that is shared by all {@link DigitalInput}s created by this provider.
+     */
+    private final ParallelPort port;
+
+    /**
+     * Factory responsible for creating digital inputs. Actual construction is fulfilled by
+     * {@link VirtualDigitalInput}. The abstraction allows composition of additional or shared
+     * responsibilities, avoiding possible inheritance from a base class to benefit from shared code.
+     */
+    private final VirtualDigitalFactory<DigitalInput, DigitalInputConfig> factory;
+
+    /**
+     * Construct a new provider which uses the given parallel port to allocate devices.
+     * @param port the underlying parallel port
+     */
+    public VirtualDigitalInputProvider(ParallelPort port) {
+        this.port = port;
+        this.factory = new ValidatingVirtualDigitalFactory<>(port.config(), VirtualDigitalInput::new);
+    }
+
+    @Override
+    public DigitalInput create(DigitalInputConfig config) {
+        return factory.createDigital(config);
+    }
+
+    /**
+     * {@link DigitalInput} implementation backed by {@link VirtualDigitalInputProvider#port}
+     */
+    private class VirtualDigitalInput
+        extends DigitalBase<DigitalInput, DigitalInputConfig, DigitalInputProvider>
+        implements DigitalInput {
+
+        /**
+         * Mask for the BCM pin associated with this digital input.
+         */
+        private final int bcmMask;
+
+        private final Map<DigitalStateChangeListener, ListenerPair> listeners = new HashMap<>();
+
+        private VirtualDigitalInput(DigitalInputConfig config) {
+            super(VirtualDigitalInputProvider.this, config);
+            this.bcmMask = 1 << config.bcm();
+        }
+
+        @Override
+        public DigitalState state() {
+            var intValue = ((port.read() & bcmMask) != 0) ? 1 : 0;
+            return DigitalState.state(intValue);
+        }
+
+        /**
+         * Hypothetical implementation of adding listeners to the digital input.
+         * <p>
+         * This simply forwards any event on the port to every digital listener regardless of whether value-change
+         * on the port actually affects the digital input.
+         * <p>
+         * If we want to filter events for each digital device, we might need to track state and compare the before
+         * and after values of the port.
+         * <p>
+         * If a port is used to back only a single pin, this is not an issue as the events would have a 1:1 mapping.
+         */
+        @Override
+        public DigitalInput addListener(DigitalStateChangeListener... listener) {
+            ParallelPort.Listener parallelPortListener = _ -> {
+                for (var l : listener) {
+                    l.onDigitalStateChange(new DigitalStateChangeEvent<>(this, state()));
+                }
+            };
+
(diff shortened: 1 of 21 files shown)


pi4j-core/src/main/java/com/pi4j/io/gpio/parallel/VirtualDigitalInputProvider.java
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * The parallel port instance that is shared by all {@link DigitalInput}s created by this provider.
+     */
+    private final ParallelPort port;
+
+    /**
+     * Factory responsible for creating digital inputs. Actual construction is fulfilled by
+     * {@link VirtualDigitalInput}. The abstraction allows composition of additional or shared
+     * responsibilities, avoiding possible inheritance from a base class to benefit from shared code.
```
stefanhaustein, line 37: Can't we do the validation in a simple plain method until there is anything that actually justifies this  indirection?

IAmNickNack: Absolutely, but I'd argue against it.

I think there are multiple aspects which could already justify the approach:

* Composition is a reasonable way of sharing functionality between the input and output providers. 
* Without copy-pasting, the alternative would be to add yet another abstract class to the hierarchy.
* Both providers are stateless factories with no mutation and pure functions. Offloading the validation doesn't only offload the logic. It also offloads the state which is required to be maintained for it to work. So , this approach also answers the question of where to maintain state when sharing code
* Unit testing is more straightforward when the test cases don't have to be concerned with constructing fakes or other responsibilities.

stefanhaustein: Just the amount of back and forth... It's named factory but it delegates to the real factory which is the provider. And because it pretends to be a factory it has to have this horrible generics salad.

In reality the only implementation just manages some state and provides validation based on the state, which would be a trivial addition to ParallelPortBase.

And do we really have to add new providers while we are trying to kill them?

It's really not a pleasure to jump between a gazillion of classes when trying to understand control flow. 

Let's not keep adding more premature abstraction anitpatterns here just because the rest of the code has them, in particular while we try to reduce unnecessary indirections elsewhere.

IAmNickNack: >  we are trying to kill them?

Not me. I've said my piece on this and I've not seen any movement towards an alternative. 

> t's really not a pleasure to jump between a gazillion of classes when trying to understand control flow.

The point is, If you know the abstraction's intention, you don't have to.

IAmNickNack: I'd also argue in this instance:

* It's technically a function reference. The fact that I've given it a name is the only thing that really makes it an abstraction.
* The "abstraction" is intentionally localised and package private. It's not polluting the API

stefanhaustein: > > we are trying to kill them?
> 
> Not me. I've said my piece on this and I've not seen any movement towards an alternative.

ProviderProvider compatibility issues have stopped my last attempt for now (v5) but this doesn't mean it's off the table for future versions and I don't think we should expand on contentious concepts, making removal even harder. I think Robert and DigitalSmile and I were clearly in favour of removal in #637 and Frank seemed to be neutral.

> > t's really not a pleasure to jump between a gazillion of classes when trying to understand control flow.
> 
> The point is, If you know the abstraction's intention, you don't have to.

That's my point. It isn't really what it pretends to be and I found the control flow with the provider injecting itself into another factory that implements a factory interface hard to follow and surprising.

IAmNickNack:
> And do we really have to add new providers while we are trying to kill them?

This comment would be more helpful and productive if it were made in a thread attached to the providers, not in the context of another discussion. 

A code review really ought to encourage and facilitate the change the reviewer would like to see. What would have been useful here would be links to the relevant issues, code examples or PRs. I don't think #702 would count because a) it's closed and b) with changes to 118 files, it's not a PR designed to be readable.

It also implies, at least to me, that the any new device types are on hold until all the code and documentation is refactored to conform to the changes that have been abandoned in #702 

Without context, or suggestions for improvement, it just comes across as unfriendly criticism.

---

I don't have the ability to request reviews, so cc'ing others instead

@FDelporte @eitch @DigitalSmile @taartspi

stefanhaustein:
> > And do we really have to add new providers while we are trying to kill them?
> 
> This comment would be more helpful and productive if it were made in a thread attached to the providers, not in the context of another discussion.
> 
> A code review really ought to encourage and facilitate the change the reviewer would like to see. What would have been useful here would be links to the relevant issues, code examples or PRs. I don't think #702 would count because a) it's closed and b) with changes to 118 files, it's not a PR designed to be readable.
>
> It also implies, at least to me, that the any new device types are on hold until all the code and documentation is refactored to conform to the changes that have been abandoned in #702

I don't think this has to be the case:

- We can add new device types without providers and treat the existing ones as legacy. With ProviderProvider being deprecated, there should be no more places where they are exposed in non-deprecated code.
- We can add new device types with providers and remove them later -- if/when we decide to go ahead with this 
 
After all, from a users perspective, it should be a (mostly) invisible implementation detail

> Without context, or suggestions for improvement, it just comes across as unfriendly criticism.

You took actively part in all or most of the providers discussions so I was assuming you were fully aware of it. 

I think when writing code for embedded, it's important to not forget about the resource constraints of devices such as a Pi zero. I used to work in a situation between factions adding "enterprise style" code to a mobile code base on one side -- and on the other side another faction promoting C because the bloat clearly proves that Java is unsuitable. I guess that has made me defensive about what seems like not technically absolutely necessary. A class here and there won't matter but in the it adds up to a lot of fat.

I am sorry that I came across so negatively.

I really like the quality of your code and the general direction of your change. My primary beef here is adding a Factory that really seems to be a validator. Perhaps you could consider potential alternatives? My preference would be to move the logic into an existing place elsewhere (base), but I'd be fine if it would be an explicit Validator. I don't like the providers but I could live with them.

> I don't have the ability to request reviews, so cc'ing others instead

@FDelporte can you fix this? 
 
> @FDelporte @eitch @DigitalSmile @taartspi

stefanhaustein:
PS I think the previous problem with ProviderProvider (IO creation bypassing registration) shows that Provider is a leaky abstraction and should (at least) be hidden from the user.

IAmNickNack:
What's a leaky abstraction?

I would agree that most of what provider does is irrelevant to the user. I think a framework should expose the concept of construction in some form though. 

The one useful function is `<T extends IOConfig, V extends IO> V create(T)`. So long as this concept is exposed somehow, it's irrelevant to the user how it's implemented or what fulfils it. 

---

I'm having trouble imagining how the context would provide construction for a virtual device which is dependent on an intermediate device, without an abstraction for a constructor operation

stefanhaustein:
> I'm having trouble imagining how the context would provide construction for a virtual device which is dependent on an intermediate device, without an abstraction for a constructor operation

Well the logical "context" / namespace for the virtual ports is the ParallelPort, so I assumed a create method there, which doesn't seem to be the case... 

Would you mind splitting this PR, moving virtual ports to a followup that's easier to digest and discuss? 

ParallelPort should be quite useful on its own already...

FDelporte:
Waiting to merge, @IAmNickNack will do a final review first.

stefanhaustein:
I'd very strongly appreciate if we could leave the controversial parts (VirtualDigitalInput/Output and the supporting infrastructure) out of the initial checkin here.

stefanhaustein (changes requested):
Logging this as a review here so it doesn't get lost: This is a big change, let's move VirtualDigitalInput/Output and supporting infrastructure to a followup.

PR #744, opened 2026-08-17
Implement defaults for missing I2C methods based on writeThenRead 

stefanhaustein:
-- or what else is available and fits, so actual implementations only need to implement writeThenRead. 

(addressing a todo in the mcp2221 bridge, which can be simplified (together with ffm)) accordingly.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/i2c/I2C.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/i2c/I2C.java
@@ -158,4 +158,62 @@ public interface I2C
      * @return the value returned by the action
      */
     <T> T execute(Callable<T> action);
+
+
+    @Override
+    default int read() {
+        byte[] ioBuffer = new byte[1];
+        read(ioBuffer, 0, 1);
+        return ioBuffer[0] & 255;
+    }
+
+    @Override
+    default int write(byte b) {
+        write(new byte[] {b}, 0, 1);
+        return 1;
+    }
+
+    @Override
+    default int readRegister(int i) {
+        byte[] ioBuffer = new byte[] {(byte) i};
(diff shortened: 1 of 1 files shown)


pi4j-core/src/main/java/com/pi4j/io/i2c/I2C.java
```diff
@@ -158,4 +158,62 @@ default I2CRegister register(int address) {
      * @return the value returned by the action
      */
     <T> T execute(Callable<T> action);
+
```
taartspi, line 161: I agree on the plan to reduce what is implemented in providers.    We need to force any i2c provider to implement   
```  java
   default void writeThenRead(byte[] writeBuffer, int writeOffset, int writeSize, int readDelayNanos, byte[] readBuffer, int readOffset, int readSize) {
  ```
The default implementation is not correct. This is meant to be a start- write-restart-read-stop, a single atomic operation.  When I tested FFM it was doing a single operation, we need to force the next implementation too do the same.    
Can this single method be marked abstract to force its implementation ?

taartspi (changes requested):

pi4j-core/src/main/java/com/pi4j/io/i2c/I2C.java
stefanhaustein: Yes, let's do this in a followup where we switch over ffm to only implement this method and make it abstract in an atomic small change

EDIT: Expanding defaults and fixing implementations can be seen as two two different steps, that's why I'd prefer to keep them separated. And cleaning up implementations is probably easier when we have all the defaults in place.

eitch (approved):

taartspi (approved):
Agree we do make my SUG item next round

PR #746, opened 2026-08-19
Push the writeThenRead method down to I2C implementations 

stefanhaustein:
Remove it from I2C, so implementations are forced to provide an implementation

For now, I am just pushing the existing code to all I2C implementations

This allows cleaning up implementations individually in clean isolated changes.

### DIFF
--- a/pi4j-core/src/main/java/com/pi4j/io/i2c/I2C.java
+++ b/pi4j-core/src/main/java/com/pi4j/io/i2c/I2C.java
@@ -86,27 +86,6 @@ public interface I2C
         return device();
     }
 
-    /**
-     * Method to write the writeBuffer, and then a read into the readBuffer
-     * in a single atomic operation.
-     *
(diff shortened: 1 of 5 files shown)


IAmNickNack:
What's the benefit of this?

stefanhaustein:
> What's the benefit of this?

Followup to #744 partially addressing Tom's request: Make clear what should be implemented by making it abstract. writeThenRead is supposed to have different bus behaviour than a sequence of a write and a read and this should be implemented in i2c providers properly. The only device I am aware of where this matters is BMP 280 in SPI mode but there it was hard to find and debug so we'd like to adjust this long term for i2c, too

taartspi:
The i2c seeproms are an example of the restart behavior expected.   We have
had forum requests to access some devices that require a number of command
bytes be written and then the restart creates the data being read.

eitch (approved):

PR #748, opened 2026-08-24
I2CFile  read register methods did not write the desired register add…

taartspi:
…ress into the chip, it only read the chip at what ever address was currently set

### DIFF
--- a/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CFile.java
+++ b/plugins/pi4j-plugin-ffm/src/main/java/com/pi4j/plugin/ffm/providers/i2c/impl/I2CFile.java
@@ -79,14 +79,16 @@ public class I2CFile extends I2CBase<FFMI2CBus> {
 
     @Override
     public int readRegister(int register) {
+        write(register) ;
         var buffer = new byte[1];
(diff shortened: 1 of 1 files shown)


taartspi:
In testing writeThenRead. The FILE implementation had bugs in the read register. The methods did not set the chips register with a write operation. It just did a read.  Fix needed but still the DIRECT implementation is the only one that correctly does the writeThenRead. I will update the provider documentation to state that.  

At the moment the build test fails. I will look into that tomorrow

taartspi:
.i2cImplementation(I2CImplementation.FILE)

With the changes in V5 Stefan asked that I confirm i2c writeThenRead works correctly.  
It does execute correctly as Implementation DIRECT.

I tried SMBUS, code tells you this request is not supported.

Tried FILE, it does not work correctly.  I started digging and saw the readRegister was not correct,
 Fixed that and integration tests fail. Could not fix that because of the mock behavior. 
 I believe some problems would be fixed if the dev was changed from i2c-dev to i2c-eeprom.   I 
 
  The question is:  Do i spend time to see if I can get more FILE function working or we change support 
  to partial SMBUS and full DIRECT support , remove FILE ?

taartspi:
while this sat, another PR did this same fix for the read() returning a length and not the read byte of data.   And the fact we are going to remove the Implementation.FILE.  I will close this.

Issue #742, opened 2026-08-17
Move `blink`, `pulse`, etc to `DigitalOutputOperations`


IAmNickNack:
Variations of `pulse`, `blink` and possibly `toggle` are operations _on_ a device, not _of_ it.

`DigitalOutput` would have clearer intent if these functions were moved to a `DigitalOutputOperations` utility.

These operations also make it essentially mandatory for any new `DigitalOutput` implementations to extend `DigitalOutputBase` and enter into the extension hierarchy. The alternative being to implement these operations anew if only implementing `DigitalOutput`.

stefanhaustein:
My plan (after my current PR) was to implement SoftwarePwm (which is basically blinking with some added parameters :) in utils which would then support these as well. 

For similar reasons: Be able to implement DigitalOutput in IO expanders, reducing the reliance of OnOffWrite

stefanhaustein:
After starting to look into SoftwarePwm, adding support there for these legacy methods seems a bit odd. 

So it would probably be better to implement DigitalOutputOperations as suggested (and perhaps deprecate the blink / pulse methods).

I am less concerned about toggle, as it should be possible to implement as a default method (if that's not the case already).

My main concern here is that the async blink methods depend on context, which I think we should avoid, so we can easily implement DigitalOutput for IO Expanders.

IAmNickNack:
> My main concern here is that the async

Personally, I don't think it should be the responsibility of the hardware implementation layer to provide asyc functionality. Especially in `core`. It could be removed from `core` entirely.

Aim to support it maybe by being thread safe, but this should be the user's responsibility to implement however they like.

If there's still feeling that it really needs to be supported, I'd add it as an extension module outside the core codebase.

eitch:
I just want to interject here, that we already have async code: All input listeners are async by nature hence we have the infrastructure for async code on the context.

IAmNickNack:
That's internal mechanics though. I'm not sure there's any benefit of leaking it out in the public API

IAmNickNack:
Is this a question of what's used or required internally and what is appropriate to expose to the user?

The module system attempts to do this by exporting only selected packages, but if any type exposed in those packages includes internal mechanics, it defeats the purpose.

There's a lot of refactoring talk going on at the moment, largely to address internal mechanics or architectural concerns. This often seems to reach the point of "user expects" or "that's exposed". 

[...]

PR #749, opened 2026-08-25
Move blink, pulse, etc to DigitalOutputOperations

stefanhaustein:
Addressing the second part of #742, simplifying DigitalOutput implementations

Operations are copied as-is, moving execution responsibility to a light executor, cutting the context dependency for non-critical tasks


stefanhaustein:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=Pi4J_pi4j&pullRequest=749) **Quality Gate failed**
> Failed conditions  [C Reliability Rating on New Code](https://sonarcloud.io/dashboard?id=Pi4J_pi4j&pullRequest=749) (required ≥ A)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=Pi4J_pi4j&pullRequest=749)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

Note that these are pre-existing issues and highlighted because code is moved around unchanged.

IAmNickNack:
> This doesn't really address the problem described in #749:

Doesn't it address the following problem (quote from the issue) by providing DigitalOutputOperations backed defaults: 

> These operations also make it essentially mandatory for any new DigitalOutput implementations to extend DigitalOutputBase and enter into the extension hierarchy. The alternative being to implement these operations anew if only implementing DigitalOutput.

To address the rest, my plan is

- Add SoftwarePwm in utils
- Deprecate the pulse / blink methods and point to SoftwarePwm as an alternative.

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputOperations.java
```diff
@@ -0,0 +1,228 @@
...
+     *                 and 5 times low for a second), also for calling the callback function.
+     *                 <p>
+     *                 If you don't want the <code>blink()</code> method to block the calling thread, pls. use the
+     *                 {@link #blinkAsync(DigitalOutput digitalOutput, int, int, java.util.concurrent.TimeUnit, com.pi4j.io.gpio.digital.DigitalState, java.util.concurrent.Callable) blinkAsync()} method instead.<br>
+     *                 <p>
+     * @param unit     The time unit used to calculate the delay.
+     * @param state    The initial state of the pin.
+     * @param callback The method to call, if any, once the blinking is done.
+     */
+    static DigitalOutput blink(DigitalOutput digitalOutput, int delay, int duration, TimeUnit unit, DigitalState state, Callable<Void> callback) {
```
IAmNickNack, line 66: If this class had a `DigitalOutput` property, these functions could be non-static and share the exact same method signature as already exists on the output interface. 

Maybe not an issue with the all-arg variants, but if there's a possibility that the user references `DigitalOutput pulse(int interval, TimeUnit unit)` as a `BiFunction` function reference and delegates the actual decision about whether to use `pulseHigh` or `pulseLow` to runtime, migrating to `Operations` be more work than necessary.

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputOperations.java
```diff
@@ -0,0 +1,228 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+/** Internal helper implementing legacy operations for DigitalOutput. */
+final class DigitalOutputOperations {
```
IAmNickNack, line 14: Rather than being a helper, this could / should be an interface which owns the spec for the operations.

The original premise of #742 was that the operations don't belong on `DigitalOutput` and should be relocated

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputOperations.java
stefanhaustein: This class is not intended to be exposed to users. It's sole point is to be able to have default implementations for all blink/pulse methods in DigitalOutput (which we should deprecate there), addressing point 2 in your issue (DigitalOutput implementation complexity without inheriting from DigitalOutputBase).

I am not fundamentally opposed to exposing the class in the form you suggest, wrapping a DigitalOutput. I just thought it provides little non-trivial value over something like SoftwarePwm (tbd), so we should rather deprecate the original methods without a 1:1 replacement than carrying these helpers over into the future in a new, slightly better form.

stefanhaustein: That said, I think we should just go ahead with this change as is, keeping PRs atomic / minimal:

- It clearly addresses a significant part of the issue
- It doesn't make anything worse
- As DigitalOutputOperations is currently not exposed, we retain freedom to change it as we like
- In which way we expose it (if at all) is probably better decided when we have SoftwarePwm as a clean replacement for some of the functionality

IAmNickNack:
> Doesn't it address the following problem (quote from the issue) by providing DigitalOutputOperations backed defaults:
> 
> > These operations also make it essentially mandatory for any new DigitalOutput implementations to extend DigitalOutputBase and enter into the extension hierarchy. The alternative being to implement these operations anew if only implementing DigitalOutput.

Technically, yes. If this point is considered in isolation. It doesn't address the preceding two points:

> Variations of pulse, blink and possibly toggle are operations on a device, not of it.
> 
> DigitalOutput would have clearer intent if these functions were moved to a DigitalOutputOperations utility.

stefanhaustein:
> > Doesn't it address the following problem (quote from the issue) by providing DigitalOutputOperations backed defaults:
> > > These operations also make it essentially mandatory for any new DigitalOutput implementations to extend DigitalOutputBase and enter into the extension hierarchy. The alternative being to implement these operations anew if only implementing DigitalOutput.
> 
> Technically, yes. If this point is considered in isolation. It doesn't address the preceding two points:
> 
> > Variations of pulse, blink and possibly toggle are operations on a device, not of it.
> > DigitalOutput would have clearer intent if these functions were moved to a DigitalOutputOperations utility.

I don't think there we have a policy here that a PR needs to fully fix an issue.

One issue why I don't want to make DigitalOutputOperations public yet is that the name is very broad. Shouldn't we make it more specific to the functionality it provides? If we add support for bit patterns we could name it DigitalOutputSequenceModulator

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputOperations.java
eitch: When we add an interface, then we need a subclass, thus needing a way to show the user how to create the class which means we need a factory/provider of sorts. DigitalOutputSequenceModulator is a mouth full. Do we see a need to implement different versions of pulse and blink? In a way it's always been something that doesn't really fit core code, no?

Having a simple version in DigitalOutputOperations is fine for me, but an interface seems wrong. Perhaps that would then make sense in pi4j-drivers?

stefanhaustein:
Ping?

eitch:
> Ping?

There were comments made, which weren't addressed thus this went forgotten =)

pi4j-core/src/main/java/com/pi4j/io/gpio/digital/DigitalOutputOperations.java
stefanhaustein: > When we add an interface, then we need a subclass, thus needing a way to show the user how to create the class which means we need a factory/provider of sorts. DigitalOutputSequenceModulator is a mouth full. Do we see a need to implement different versions of pulse and blink? In a way it's always been something that doesn't really fit core code, no?

One reason to have/keep it in core is the existing dependency from core

> Having a simple version in DigitalOutputOperations is fine for me, but an interface seems wrong. Perhaps that would then make sense in pi4j-drivers?

I agree. I think we tend to have this kind of thinks in utils. I don't think drivers would be the right place as it doesn't add support for additional hardware...

In any case, I think we should move ahead with the change as-is as it improves the subclassing situation and leaves everything else open for a followup, as nothing new is exposed....

stefanhaustein:
> > Ping?
> 
> There were comments made, which weren't addressed thus this went forgotten =)

I think they should be all addressed now?

eitch (approved):

PR #753, opened 2026-09-18
FFM I2C: return read() unsigned in the direct and file implementations

### LINKED ISSUE #752: FFM: I2CDirect.read() and I2CFile.read() return the byte sign-extended (follow-up to #594)
**Version:** 4.0.2, also current `main` (`I2CDirect.java` line 146, `I2CFile.java` line 60)
**Plugin:** pi4j-plugin-ffm

### What happens

`I2C.read()` returns the raw `byte` widened to `int`, so values 0x80–0xFF come back negative. A PCF8574 with all pins high answers 0xFF and `read()` returns **-1**, which by the `read()` contract means "nothing read". The default implementation is affected (DIRECT is chosen when no `i2cImplementation` is set), and so is FILE; SMBUS is correct.

```java
// I2CDirect
public int read() {
    return internalRead(new byte[1], 0, 1)[0];          // sign-extended
}
public int readRegister(int register) {
    return Byte.toUnsignedInt(internalRead(...)[0]);    // fixed in #598
}
```

`I2CFile.readRegister(int)` has a related slip: it returns `read.length` (always 1) instead of the byte value.

The unit test currently locks the bug in: `I2CTest.testReadDirect` asserts `assertEquals((byte) 0xff, direct.read())`, i.e. -1, while `testReadSMBus` asserts `0xff` for the same call on the SMBus implementation.

### How to reproduce

Any device whose first byte has bit 7 set; a PCF8574 (pins idle high) is the simplest:

```java
Context pi4j = Pi4J.newAutoContext();
I2C pcf = pi4j.create(I2C.newConfigBuilder(pi4j).id("pcf").bus(1).device(0x20).provider("ffm-i2c").build());
System.out.println(pcf.read());   // -1; with .i2cImplementation(I2CImplementation.SMBUS): 255
```

Seen in practice as an `i2cdetect`-style scan that skipped the PCF8574 on a Waveshare Pioneer600 (treated `read() < 0` as no acknowledge) while listing the DS3231 and PCF8591 next to it, whose first bytes happened to be below 0x80.

### Expected

`read()` returns 0–255 in all three implementations, as `I2CSMBus.read()` and `I2CDirect.readRegister(int)` already do; `I2CFile.readRegister(int)` returns the value rather than the count.

### Prior art

- #594 / #598: the same defect in `I2CDirect.readRegister(int)`, fixed with `Byte.toUnsignedInt` (Jan 2026). `read()` was not touched.
- #164: the same defect in the mock plugin's single-byte read (2021).

PR with the two one-line fixes, the `readRegister(int)` return value and updated unit tests follows.

mstahv:
I2CDirect.read() and I2CFile.read() returned the raw byte widened to int, so 0x80..0xFF came back negative and a PCF8574 with all pins high (0xFF) read as -1, which the I2C.read() contract reserves for "nothing read". The SMBus implementation and, since #598, I2CDirect.readRegister(int) already use Byte.toUnsignedInt; this applies the same to read(). I2CFile.readRegister(int) also returned the number of bytes read instead of the value; it now returns the unsigned byte like the other implementations.

The unit test for the direct read asserted the sign-extended -1; it now expects 0xFF, and the file read test uses bytes with the top bit set so both paths are covered.

Closes #752


stefanhaustein (approved):
Thanks! @DigitalSmile seems to be busy and this looks like a very clear & straightforward case, so I took the liberty to stamp this O:)

